PR #3, opened 2018-09-20
Initial work on a macro to reduce boilerplate code

jcward:
Hey @larsiusprime, here's a macro to reduce boilerplate code of calling the scripts. Let me know if you have any questions, it should be pretty readable code, and the demo still works just as before.

Each script invocation sets up two variables: `script_result` and `script_error` -- I think there is sample usage of both. If a function body exists, the script is prepended. The scope variables can be defined in the `@:hscript` metadata - (todo: we could also allow `*` as a shortcut to give the script access to all the classes members instead of having to write them all out).

My various notes from the email still apply:

- it seems like there can only be a single scriptable function of each name in the whole project.
- it will load assets every time a scriptable class is instantiated

In the openfl_hscript sample, it's only the top-level class that's scriptable, so neither of those restrictions is much of a problem. But you could imagine making many classes scriptable, so those concerns would magnify.


jcward:
Oh, I suppose the `Assets.getText()` calls [injected into the constructor](https://github.com/larsiusprime/polymod/pull/3/files#diff-6d436f3a381c7fe62e4deb89959b912cR81) are specific to OpenFL, you'll probably want to factor out framework-specific ways to load the scripts.

larsiusprime:
This is great stuff, Jeff! REALLY impressed. I updated the documentation with my best understanding of this, please let me know if I got anything wrong?

https://github.com/larsiusprime/polymod/blob/master/README.md#scripting

PR #18, opened 2019-10-22
Add ability to create parse rules for specific files

mrcdk:
This way, you can add specific rules for specific files like:
```haxe
		var specificRule:CSVParseFormat = new CSVParseFormat(";", false);
		parseRules.addFile("data/myspecificfile.csv", specificRule);
```


larsiusprime:
Looks good!

PR #25, opened 2021-07-15
make compatible with haxe 4.2.1

darmie:


prokube (approved):
that `ParseRules` diff reminds me of before I used GitHub Desktop and I just cut-pasted the entire edited file in LOL

I'm assuming that `ParseRules` is fine, so yeah it gets an approval from me

larsiusprime:
Thanks, just fixed this.

PR #28, opened 2021-09-03
Remove the errors of polymod

Cheemsandfriends:
This PR removes the "XMLMerge.hx:189: characters 10-23 : Warning : This typedef is deprecated in favor of haxe.xml.Access" typedef.
(It's a minor change but I guess it's good so you don't have that error lol)


larsiusprime:
Thanks, just fixed this.

PR #31, opened 2021-09-14
Fix the extensionMap parameter.

MasterEric:
If extensions is null, we already have code to check for that and initialize it here:

https://github.com/larsiusprime/polymod/blob/bb5f0a120419ac3a7132d96aff1e6f7a36b97d67/polymod/backends/PolymodAssetLibrary.hx#L262

If extensions is not null (i.e. it was provided by the user) then re-initializing it would ignore the user's preferences.


larsiusprime:
Thanks, just fixed this.

PR #48, opened 2021-09-23
Remove deprecation warnings from `Std.is`

Geokureli:


MasterEric:
Already fixed in version 1.3. Thanks for contributing though :)

Issue #59, opened 2021-11-18
Allow for configuring metadata file names


MasterEric:
Allow the names of these metadata files to be redefined at compile time using a `project.xml`:

* `_polymod_meta.json`
* `_polymod_icon.png`
* `_polymod_pack.txt`

MasterEric:
This is done now btw

PR #84, opened 2021-12-29
Fixed error not using correct Mod Metadata File

DotWith:
Previously the modMetadata wasn't being used for this error. This [issue ](https://github.com/larsiusprime/polymod/issues/59) should of been closed by now!


MasterEric:
Nice catch!

Issue #85, opened 2022-01-05
[Some sort of Enhancement?] implement an automated reader of folders in assets


Cheemsandfriends:
Hello!
I know it sounds kinda stupid but, it would be great that there should be a reader of folder in assets to have and that would be the assetlibrarypaths in case that the person who uses Polymod and is lazy, doesn't have to insert the folder manually one by one

MasterEric:
I don't understand the prompt.

Adding support for additional asset libraries (a planned feature now btw) would require defining them manually. If you automatically added new asset libraries based on the folder names, it would break mods trying to replace default assets.

Additionally, having assets automatically split into separate libraries would be confusing for developers who are only using one asset library.

Cheemsandfriends:
@MasterEric, I know that additional asset libraries are a feature that is planned, but there is something that I can't take and it's adding several folder names if you have like a ton. I thought that just reading the folder names if the dev doesn't want to add some stuff sounded like a nice idea!
So like it's the name of the folder and the dir would add at the beginning `./` cos it's the main thing.

Idk maybe it's not a good idea.

PR #87, opened 2022-01-07
Add a if statement so that you don't need to add the assetLibraryPaths unless you are crazy

Cheemsandfriends:
Hello, As I said in the Issue/trytoEnhancementthing number #85, I said that it would be great to add an automated thing to load your stuff. This adds the thing and pretty much it works like the old Polymod imo but even better

### DIFF
--- a/polymod/Polymod.hx
+++ b/polymod/Polymod.hx
@@ -25,6 +25,7 @@ package polymod;
 
 import polymod.fs.IFileSystem;
 import haxe.Json;
+import haxe.io.Path;
 import haxe.io.Bytes;
 import polymod.util.SemanticVersion;
 import polymod.util.Util;
--- a/polymod/Polymod.hx
+++ b/polymod/Polymod.hx
@@ -166,6 +166,20 @@ class Polymod
 			new polymod.fs.StubFileSystem();
 			#end
 		}
+		if (params.frameworkParams == null || params.frameworkParams != null && params.frameworkParams.assetLibraryPaths == null)
+		{
+			var modStuff:Map<String, String> = [];
+			var foldersOnly = Lambda.filter(fileSystem.readDirectory("./assets"), _ -> 
+			{
+				return (Path.extension(_) == "");
+			});
+			for (i in 0...foldersOnly.length)
+			{
+				modStuff[foldersOnly[i]] = './${foldersOnly[i]}';
+			}
+
+			params.frameworkParams = {assetLibraryPaths: modStuff};
+		}
 
 		if (params.modVersions != null)
 		{
--- a/polymod/Polymod.hx
+++ b/polymod/Polymod.hx
@@ -152,35 +152,22 @@ class Polymod
 
 		var modMeta = [];
 		var modVers = [];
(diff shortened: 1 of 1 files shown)


Cheemsandfriends:
This is in case that someone is lazy (Me!!!) and doesn't want to be doing all that fuss of "folder name" => "./folder name"

TheDisabledKid:
can you add that to the polymod that you forked?

TheDisabledKid:
also i did what you did and i got this error /usr/local/lib/haxe/lib/polymod/git/polymod/Polymod.hx:164: characters 89-93 : Type not found : Path

Cheemsandfriends:
> also i did what you did and i got this error /usr/local/lib/haxe/lib/polymod/git/polymod/Polymod.hx:164: characters 89-93 : Type not found : Path

Oh oops. I forgot the import thing
Wait a minute

Cheemsandfriends:
@TheDisabledKid, I guess that now there isn't any errors now. If there is I forgot the path of importing the path but I think I'm correct lol

TheDisabledKid:
ok ima test it to see if it works

MasterEric:
This doesn't really make sense.

First off, this logic messes up with the HEAPS and Kha backend. Remember, they don't use Lime.

Secondly, the list of folders that are excluded is hardcoded. As it stands, a folder named `img` would get turned into an asset library.

The primary thing to remember is that having multiple asset libraries is an EDGE CASE that FNF happens to use. The default behavior of most Lime and Flixel applications is to use a single asset library `default`.

The current/planned behavior maintains compatibility with existing code and minimizes hassle in the default case, while allowing for intuitive and well-defined behavior in the edge case, while also maintaining support for other engines. I don't see a need to interfere with that.

I appreciate your attempts to contribute, but I'm going to close this change.

Cheemsandfriends:
Hi @MasterEric!

I wanted to say that if you don't like the PR, it's Ok to do that, but at least please make a feature so it really isn't necessary to use the goddamn thing, if you don't use it, either it crashes or the mod doesn't even work.

> First off, this logic messes up with the HEAPS and Kha backend. Remember, they don't use Lime.

I thought that it was obligatory to use the thing only in Lime, which it says it on the description of the thing. I'll add like a small thing to conclude that makes `if params.framework == LIME` thing. and then I'll check up how it works on different stuff. and I'll maybe add more stuff on the PR when I have an idea about it

> The current/planned behavior maintains compatibility with existing code and minimizes hassle in the default case, while allowing for intuitive and well-defined behavior in the edge case, while also maintaining support for other engines. I don't see a need to interfere with that.

I don't want to be rude but it's a huge pain in the ass setting the customPathLibraries if you have a lot of libraries included. It's not a big deal to have like one or two folders in the assets thing. But imagine if the dev has like 15 or 20 folders and starts using Polymod. if you waaant support, add support for lazy devs that also want to just not completing the custom asset libary paths thing pls.

Please don't take this as a hated message towards Polymod cos you didn't accept this PR, I love the idea and it's one of the greatest thing I've ever seen! but the Asset Paths should be fixed the way it should work even if you don't set it. like it automatically sets it for you

PR #93, opened 2022-01-19
fixed polymod

ThatDisabledKid:
I have fixed the polymod


ThatDisabledKid:
@MasterEric come and look at this

MasterEric:
I'm working on a 1.4.1 that fixes this but I need to fix another separate crash issue first.

This is not the correct line, by the way. The proper fix is to change `openfl.Assets.cache.cache.removeBitmapData` to simply `openfl.Assets.cache.removeBitmapData`

ThatDisabledKid:
mk let me fix that real quick

PR #105, opened 2022-02-22
added a question mark

beeanie:
idk if this helps


MasterEric:
Actually this might need an extra line but it's a good idea

PR #109, opened 2022-03-11
made it public

McJonnycat:
i just made it a public static function


MasterEric:
The values of PolymodConfig are properties, not variables. The `get_debug` function gets automatically called when you access `PolymodConfig.debug`, which is a public value.

> An accessor method (or short accessor) for a field named field of type T is a getter named get_field of type Void->T or a setter named set_field of type T->T.

The getter function is specifically NOT supposed to be public. Just access the value normally.

McJonnycat:
oh my bad

PR #118, opened 2022-04-28
made it !=null instead of ==null

McJonnycat:


MasterEric:
This doesn't make any sense, the code is looking to define the value if it is not already defined (i.e. if it is null).

PR #125, opened 2022-06-10
fixed the hscript errors

McJonnycat:
haha i did it


McJonnycat:
and also getting the git version of hscript does not work they both give out the same error

McJonnycat:
@MasterEric Please merge this

MasterEric:
Your solution would break line numbers on error messages rather than actually fixing the problem.
I already committed a PROPER bug fix for the `hscriptPos` errors to `develop` two days ago.

PR #142, opened 2022-11-24
Added support for compressed zip files in ZipFileSystem

UncertainProd:
This PR adds support for reading compressed zip files as well to the ZipFileSystem classes
(shout out to haxe.zip.InflateImpl, which I didn't realize was already cross-platform lol)


MemeHoovy (approved):
dope

MasterEric:
Does reading individual files from a compressed ZIP require loading the entire ZIP into memory?

UncertainProd:
On sys targets it doesn't seem to need it, as it just seeks to that point in the zip file and reads the bytes. But since html5 doesn't have a way of seeking through files, it ends up storing the entire zip in memory for the web.
Although, I'm not sure how to inspect how much of the zip is loaded into memory, do you know of any way?

MasterEric:
I was thinking you'd assign the program a very large ZIP then test it like
that, but since it's using seek on Sys platform it probably does work
correctly

UncertainProd:
Looks like it doesn't load the entire zip, I put some large images along with some small ones into one of the zips and only tried to load the small one and it didn't seem be affected by how long it would have taken to load the large image

MasterEric:
Makes sense. Were you looking at duration or memory usage there?

UncertainProd:
Duration

MasterEric:
I merged this after some cleanup but it appears Github doesn't understand that, guess I'll just close this.

Good work :+1:

PR #147, opened 2023-01-11
Different fixes for feature/blacklist

AltronMaxX:
Maybe not all of them are good, but without them this version of polymod don't work


MasterEric:
Hey, this branch is work-in-progress. I haven't actually gotten around to making sure the the blacklist/alias stuff works as expected.

Also I have like 12 local commits yet to be pushed lol, those probably have all the fixes you were looking at.

image

PR #154, opened 2023-01-19
Null object reference fix for scripted functions

AltronMaxX:
It`s appears only when you using scripted functions, because PolymodInterpEx creates with null proxy.


MasterEric:
Looks good to me

PR #169, opened 2023-08-01
Compile fix when not using `hscript`.

MAJigsaw77:
Adding `hscript` define were `PolymodScriptClass` is mentioned.


EliteMasterEric (approved):
thanks

PR #174, opened 2024-07-06
Update PolymodAbstractScriptClass.hx to have "findVar" in resloveField function

7oltan:
i wanted to add this to help with my code of making safe scriptSet and safe scriptGet (soft coded) which is basically:
get variable if it doesnt exist return default value
but i cant do that because its not possible to check for a varaible so i added this change

thank you!!


EliteMasterEric:
Passing a `VarDecl` outside the hScript class is un-intuitive and risks breaking stuff.

Maybe instead a `hasVar` function which returns a boolean if `findVar` is not null would make more sense here?

7oltan:
Both work, tho I used "hasFunction" and it works but idk about the hscript stuff it scares me

Issue #185, opened 2024-11-10
[HashLink] Local functions cause errors


NotHyper-474:
I believe the issue lies somewhere in this section that extends/overrides HScript's implementation as removing it solves this issue.
https://github.com/larsiusprime/polymod/blob/0fbdf27fe124549730accd540cec8a183f8652c0/polymod/hscript/_internal/PolymodInterpEx.hx#L234-L360

Repro code:
```haxe
var func = (t:FlxTimer)-> {
   trace('test');
};

new FlxTimer().start(1.0, func); // Causes an "Access Violation" error 
func(null); // Causes a "Can't call closure with value" error
```

Using develop branch, have not tested with 1.8.0. On the C++ target and etc it works just fine.

NotHyper-474:
Seems like what's causing this issue is the fact that `newFun` is being assigned to itself, HashLink doesn't seem to like that for some reason. So doing the following solves the error for me:
```diff
-				newFun = Reflect.makeVarArgs(newFun);
+				var resultFun = Reflect.makeVarArgs(newFun);
 				if (name != null)
 				{
 					if (depth == 0)
 					{
 						// Store the function as a global.
-						variables.set(name, newFun);
+						variables.set(name, resultFun);
 					}
 					else
 					{
 						// function-in-function is a local function
 						declared.push({n: name, old: locals.get(name)});
-						var ref = {r: newFun};
+						var ref = {r: resultFun};
 						locals.set(name, ref);
 						capturedLocals.set(name, ref); // allow self-recursion
 					}
 				}
-				return newFun;
+				return resultFun;
```

NotHyper-474:
Turns out it's some bug with how HashLink is handling implicit types. `newFun` is of type `function(Array<Dynamic>)` and a value of type `Dynamic` is being assigned to it and that's what's causing the crash for some reason. To workaround this you can simply make the type of `newFun` explicitly as Dynamic: `var newFun:Dynamic = function(args:Array<Dynamic>)`.

Gonna make a separate issue upstream.

PR #187, opened 2024-12-26
Fix error/crash with local functions on HashLink

NotHyper-474:
See #185


NotHyper-474:
Upstream issue has been closed/resolved (https://github.com/HaxeFoundation/hashlink/issues/735) but we won't see it until Lime updates to HL 1.15 so I'll keep this open.

PR #194, opened 2025-02-22
[FEATURE] Import Scripted Classes

lemz1:
## DESCRIPTION
This pr adds the ability to import scripted classes like you would do in regular haxe, meaning:
```haxe
package my.mod.game;

import game.ScriptedGame;
import my.mod.objects.ModdedPlayer;

class ModdedGame extends ScriptedGame {
  public function new() {
    var player = new ModdedPlayer();
  }
}
```

## EXPLANATION
We create another imports variable to the `PolymodClassDeclEx`, because a script might import a class that hasn't been registered yet.
After all classes have been registered and parsed, we validate the imports. If the import still can't be found we give the user an error.

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -68,6 +68,13 @@ class PolymodInterpEx extends Interp
 		if (clsRef != null) return clsRef.instantiate(args);
 
 		@:privateAccess
+		if (getClassDecl().imports != null && getClassDecl().imports.exists(cl)) 
+		{
+			var clsRef = PolymodStaticClassReference.tryBuild(getClassDecl().imports.get(cl).fullPath);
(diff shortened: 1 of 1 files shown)


AbnormalPoof (approved):
Yes please

cyn0x8:
will #201 replace this? or is it a separate thing
i see "ability to extend scripted classes" but idk if this implies the ability to import or not

lemz1:
> will #201 replace this? or is it a separate thing i see "ability to extend scripted classes" but idk if this implies the ability to import or not

#201 does implement importing scripted classes. So i guess #201 would also replace this pr.

EliteMasterEric:
When you want to
test a PR
but Github said
1 billion conflicts

EliteMasterEric (changes requested):
Still has merge conflicts with the experimental branch.

PR #195, opened 2025-02-22
[FEATURE] Abstract Statics Access

lemz1:
(Renamed the branch, which forced me to reopen the pr?)

## DESCRIPTION
This pr adds the ability to access the static members of abstracts.

## NOTE
`FlxColor` is the only thing that has been tested, so I don't know if other abstracts work.

## EXPLANATION
We loop through all existing abstracts, and get the variables. 

To filter out non static member variables, I have decided to exclude any variable that has a getter or setter. I don't like doing it this way, but I haven't found a better way yet.

Then we define a class that contains getters that return the value of the static in the abstract. The reason we do this, is because abstract static member variables aren't stored in the reflection table.

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -8,6 +8,8 @@ import polymod.hscript._internal.PolymodExprEx;
 import polymod.hscript._internal.PolymodClassDeclEx.PolymodClassImport;
 import polymod.hscript._internal.PolymodClassDeclEx.PolymodStaticClassReference;
 
+using StringTools;
+
 /**
  * Based on code by Ian Harrigan
  * @see https://github.com/ianharrigan/hscript-ex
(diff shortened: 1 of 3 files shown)


KoloInDaCrib:
> (Renamed the branch, which forced me to reopen the pr?)

iirc the logic behind renaming a branch is
1. create a new branch with a desired name
2. move changes from the old one
3. delete the older branch (and with that close the prs associated with it)

EliteMasterEric:
There are a LOT of merge conflicts with this one, might have to wait until later.

EliteMasterEric:
It appears this works with static functions, but not static variables somehow?

```haxe
  public static function testAbstract() {
    var color = FlxColor.fromString("RED");
    trace('FlxColor.fromString("RED") = ' + color);
    var colorHex = FlxColor.toHexString(color);
    trace('FlxColor.toHexString(color) = ' + colorHex);
    var color2 = FlxColor.BLUE;
    trace('FlxColor.BLUE = ' + color2);
  }
```

results in

```
hscriptClass(assets/gameplay/scripts/modules/EricsTestModule.hxc):81: FlxColor.fromString("RED") = -65536
hscriptClass(assets/gameplay/scripts/modules/EricsTestModule.hxc):83: FlxColor.toHexString(color) = 0xFFFF0000
hscriptClass(assets/gameplay/scripts/modules/EricsTestModule.hxc):85: FlxColor.BLUE = null
```

EliteMasterEric (changes requested):
The code itself looks good in general, if we can get my example snippet working then I 100% want to merge this.

lemz1:
I've done some testing, and it seems to sometimes work and sometimes fail?

I dont know how i managed to add RNG to code execution

EliteMasterEric:
You're right, it does seem to work but only sometimes?

Since it should work fine after a clean build I'll merge it, we can figure out what might cause it later (it probably has to do with the macro and caching data between builds).

PR #200, opened 2025-02-28
[FEATURE] Add a way to download dependencies

lemz1:
## DESCRIPTION
This pr adds functions to download the dependency using a github repo and commit hash. You can also just get the bytes of the generated zip, so that you can handle downloading it yourself.

### DIFF
--- a/polymod/Polymod.hx
+++ b/polymod/Polymod.hx
@@ -817,12 +817,17 @@ typedef ModContributor =
 };
 
 /**
- * A type representing a mod's dependencies.
- *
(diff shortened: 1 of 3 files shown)


EliteMasterEric:
Is there anything preventing this from being used to download a virus? Sounds like a security concern.

lemz1:
To be fair any mod that you would download is a security concern.

The security measures should be handled by the developer that wants to add automatic dependency downloading for their game. For example, the developer could make a popup appear that asks for permission to download not found dependencies.

PR #201, opened 2025-03-01
[FEATURE / REWORK] New Implementation For Scripted Classes

lemz1:
## NOTE
I think everything works, but I might have missed something. So feel free to tell me about issues that have popped up.

## DESCRIPTION
This pr reworks the implementation for script classes. The changes include:

- use macro `buildScriptImpls` to determine which classes can be scripted (filter can either be: `my.package`, `my.package.Module`, `my.package.Module.SpecificClass`). the macro is used inside the `project.xml`, `build.hxml`, or whatever build system you use.
- ability to extend scripted classes
- abitily to create classes that dont extend anything
- no more scriptCall, scriptSet, etc.
- custom constructors
- static variables and functions

## @:noCast
@:noCast is metadata that you can attach to an argument when calling a function. Scripted classes are automatically casted to their non scripted super class (if it exists), when parsing them to a function from a non scripted class. The reason this is done, is so that it is easier to parse scripted classes into functions from regular classes (See the example where we use `add` from FlxTypedSpriteGroup, without auto casting we would need to do: `add(obj.castType(FlxSprite))`).

Automatic casting is only done in functions, which have a definition in a regular class. This means when we override a function from a regular class, it will do auto casting. However, if we create a function, which doesn't exist in a regular class that we extend, no auto casting will be done.

@:noCast forces polymod to use the `PolymodScriptClass` instead of casting to the super class.

@:noCast is only available in function arguments.

<details>
<summary>Code</summary>

```haxe
import flixel.group.FlxTypedSpriteGroup;
import flixel.FlxSprite;

class CoolObject extends FlxSprite
{
  public function new()
  {
    super();
  }
}

class CoolGroup extends FlxTypedSpriteGroup
{
  public function new()
  {
    super();
    var obj = new CoolObject();
    add(obj); // obj will be auto casted to FlxSprite

    print(obj) // won't be auto casted, since this is purely a scripted function (it doesn't override a function from FlxTypedSpriteGroup)

    trace(obj); // trace doesnt auto cast: "PolymodScriptClass(CoolObject)"
  }
  
  public function print(o)
  {
  
  }
}

```

</details>

## EXAMPLE
<details>
<summary>CoolObject.hxc</summary>

```haxe
package mod1;

import flixel.FlxSprite;

class CoolObject extends FlxSprite
{
  var time:Float = 0;

  public function new(color:Int)
  {
    super();
    makeGraphic(1280, 720, color);
  }

  public function update(elapsed:Float):Void
  {
    super.update(elapsed);
    time += elapsed;

    y = (360 - height / 2) + Math.cos(2 * time) * 60;
  }

  public function offsetTime(value:Float):Void
  {
    time += value;
  }
}
```

</details>

<details>
<summary>CoolerObject.hxc</summary>

```haxe
package mod2;

import mod1.CoolObject;
import flixel.group.FlxTypedSpriteGroup;
import flixel.FlxSprite;

class CoolerObject extends CoolObject 
{
  var offsetTimer:Float = 0.0;

  public function new()
  {
    super(0xff00ff00);
    setGraphicSize(640, 360);
    screenCenter();
  }

  public function update(elapsed:Float):Void
  {
    offsetTimer += elapsed;

    if (offsetTimer >= 1.0)
    {
      offsetTimer = 0;
      offsetTime(1.25);
    }

    super.update(elapsed);

    x = (640 - width / 2) + Math.sin(2 * time) * 120;
  }

  public function offsetTime(value:Float)
  {
    super.offsetTime(2.0 * value);
  }
}

class CoolGroup extends FlxTypedSpriteGroup
{
  var obj:CoolerObject;
  public function new()
  {
    super();
    obj = new CoolerObject();
    add(obj);
  }
}
```

</details>


TheRealJake12:
really cool stuff

AbnormalPoof:
This PR is really interesting! The example scripts you shown have classes that extend something. Is it possible to use classes that don't extend anything?

lemz1:
> is it possible to use classes that don't extend anything?

yes

EliteMasterEric:
This one seems elaborate and will probably take a lot of time to review. I want to understand everything before I merge because it seems like a major overhaul and I'm cautious of that for obvious reasons.

One thing I can say, while scriptCall, scriptSet, etc are no longer needed, some scripts may use them and if it's not significant work to implement them they can stay for backwards compatibility.

lemz1:
> This one seems elaborate and will probably take a lot of time to review. I want to understand everything before I merge because it seems like a major overhaul and I'm cautious of that for obvious reasons.

If you have any questions you can just hit me up on discord (**lemz1**). Of course, you can also ask them here, but i might respond a bit later, since the notifications seem to not work properly for me.


> One thing I can say, while scriptCall, scriptSet, etc are no longer needed, some scripts may use them and if it's not significant work to implement them they can stay for backwards compatibility.

noted, will work on re-adding them for backwards compatibilty

EliteMasterEric:
Aside from being super outdated, I'm not sure this PR is even necessary at this point, with most of the improvements being reimplemented in other, less disruptive PRs.

What important stuff does this do now? Extending other scripted classes, extending nothing, not requiring scriptCall in scripts is all done.

NotHyper-474:
I think the only thing left would be the ability to extend almost any classes without requiring to manually implement the `HScriptedClass` interface and metadata, that would be pretty neat.

PR #204, opened 2025-03-06
Case-insensitive ZIP loading

NotHyper-474:
[This mod](https://gamebanana.com/mods/578445) shows an interesting problem that is, the modder might not care enough for the letter case of the files, which may cause inconsistent behavior (files not loading) not only in case-sensitive platforms (like Linux) but also when trying to load their mod as a ZIP.

This PR aims to solve at least the ZIP part. I might look into the Linux issue later.


NotHyper-474:
Tested with a modified version of QT Rewired (so it uses hxvlc instead) and the mod mentioned above and it seems to work pretty well.

EliteMasterEric:
This code seems good but I wonder if the behavior could be made optional (via a compilation flag or similar) since some applications may expect case-sensitive asset loading. Or maybe we could do the opposite and ensure case-sensitive asset loading on Windows.

NotHyper-474:
Done. Now you need to set a `POLYMOD_ZIP_INSENSITIVE` define to enable it (though maybe it should be enabled by default?)

> Or maybe we could do the opposite and ensure case-sensitive asset loading on Windows.

Definitely would break a mod or two (I know for sure QT: Rewired would be among these as it has a couple of vocals that don't load on Linux). Personally I'm more inclined towards ensuring case-insensitive loading on Linux instead, but up to you.

PR #209, opened 2025-04-13
[BUGFIX] Prevent Script Values being dropped early

KoloInDaCrib:
Fixes https://github.com/FunkinCrew/Funkin/issues/4158

## Description
This PR utilizes cloning to clone `PolymodInterpEx` class and to use that clone when running functions. That way, `locals` don't get cleared, since this is a new instance of the interp.

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -559,7 +559,6 @@ class PolymodInterpEx extends Interp
 		{
 		#end
 			// These overrides are used to handle specific cases where problems occur.
-
 			case EVar(name, type, expression):
 				// Fix to ensure local variables are committed properly.
 				declared.push({n: name, old: locals.get(name)});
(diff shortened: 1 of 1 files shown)


EliteMasterEric (changes requested):
Was going to test this but it has merge conflicts.

T5mpler:
Seeing as this seems expensive, it's probably best to either be able to cache this, or destroy the clone after it's done being if (but I'm not exactly sure if this would be easy to check however)

EliteMasterEric (changes requested):
This ended up with conflicts again.

T5mpler:
Tested the PR and it works. From my experience there's barely to 0 memory issues, nor are there any FPS issues (hopefully it's not because I have a good PC or anything, others are free to try).

I made my own implemention of this that uses a clone instance and simply clears and re-caches the variables when the cloning process happens to prevent any potential issues/make this process less expensive.

Overall if this PR could be looked over and further merged that would be a life saver

image

image

Starexify (approved):
Works with no issues for me, veri gud 👍

NotHyper-474:
I've seen concern about the impact of this cloning and to me it appears to stem from a misinterpretation of what `EFunction` is.
It actually refers to local functions, e.g `() -> {}`, so **if** it _**is**_ expensive it would only ever affect them.

NotHyper-474:
Pls Ashnormal I need this, my static local functions are kinda contextless

AbnormalPoof (approved):
Works great and doesn't break anything! Tested using the example scripts in the issue. The implementation is a bit questionable, but I don't think there's an easier way of doing this without being jank.

PR #220, opened 2025-07-06
[FIX] better macro checks for abstract static access

lemz1:
When a member variable has a getter or setter, we now check if it has 0 arguments, which is the case for static getters,
or 1 argument, which is the case for static setters.

### DIFF
--- a/polymod/hscript/_internal/PolymodScriptClassMacro.hx
+++ b/polymod/hscript/_internal/PolymodScriptClassMacro.hx
@@ -172,57 +172,108 @@ class PolymodScriptClassMacro {
 					var abstractImplPath = abstractType.impl.toString();
 					var abstractImplType = abstractType.impl.get();
 
-					var excludes:Array<String> = [];
 					for (field in abstractImplType.statics.get()) {
-						switch (field.type) {
-							case TFun(_, _):
-							default: 
-								continue;
-						}
+						switch (field.kind) {
+							case FVar(read, write):
+								trace(field.name, read, write);
 
-						// exclude anything that has a getter or setter
-						// most of the time i think variables that have them are not static
-						// hopefully that's true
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
What specific issue does this resolve? Do you have a minimal script that shows off the problem before/after this PR?

lemz1:
> What specific issue does this resolve? Do you have a minimal script that shows off the problem before/after this PR?

lets say we have this abstract:
```haxe
package;

abstract TestAbstract(Int) from Int to Int {
  public static var TEST_VALUE(get, never):TestAbstract;

  static function get_TEST_VALUE():TestAbstract {
    return 1000;
  }
}
```

before if we try to access TEST_VALUE in a script, it would not work, because we skip every field that has a getter or setter.

now instead we check the amount of arguments a getter or setter has.
for static fields it will be 0 for a getter and 1 for a setter
for instance fields it would be 1 for a getter and 2 for a setter (in both cases the extra argument is just a "this" parameter)

in other words this pr fixes static fields being skipped if they have a getter or setter.

this would be the script:
```haxe
import flixel.FlxBasic;
import TestAbstract;

class LolCool extends FlxBasic {
  public function new() {
    super();
    trace('LOL');
    trace(TestAbstract.TEST_VALUE); // before pr this prints "null", after this pr it prints "1000" (the actual value)
  }
}
```

If you want to test this yourself dont forget to change the if checks in the PolymodScriptClassMacro file, because we only allow "flixel.util.FlxColor" currently, so you would need to change it to check for "TestAbstract"

EliteMasterEric:
I understand now! With that context, the code makes a lot more sense.

I reviewed it and the code looks good, will need to test locally though to make sure it functions as expected.

After that, we can do more testing to make sure all functionality of `FlxColor` works as expected, then remove the restriction and test on other abstracts, and at that point abstract access should be fully functional.

PR #227, opened 2025-07-10
[FIX] Only include class in metadata if it has final fields

### LINKED ISSUE #226: [experimental branch] C++ code generated from `PolymodFinalMacro` causes certain compilers to hang
This is very annoying as a Linux user as I'm forced to use clang instead, and even it complains about the generated code
```
./src/polymod/hscript/_internal/PolymodFinalMacro.cpp:180:56: warning: stack nearly exhausted; compilation time may suffer, and crashes due to stack overflow are likely [-Wstack-exhausted]
```

And the generated code in question is obviously problematic, containing a hell of nested functions contained in a humongous 500k character line:

Image

NotHyper-474:
## Linked Issues
Closes #226 

## Description
The macro would include every single class into the metadata, making the generated C++ code massive which causes some compilers to completely hang. This mitigates the issue by making it only include classes that contains final fields.

### DIFF
--- a/polymod/hscript/_internal/PolymodFinalMacro.hx
+++ b/polymod/hscript/_internal/PolymodFinalMacro.hx
@@ -42,6 +42,8 @@ class PolymodFinalMacro
               finals.push(field.name);
             }
 
+            if (finals.length == 0) continue;
+
             var entryData = [
               macro $v{classPath},
               macro $v{finals}


lunacynne (approved):

polymod/hscript/_internal/PolymodInterpEx.hx
```diff
@@ -322,9 +322,9 @@ class PolymodInterpEx extends Interp
 									for (imp in _proxy._c.imports)
 									{
 										if (imp.name != id0) continue;
-										var finals = PolymodFinalMacro.getAllFinals().get(imp.fullPath) ?? [];
+										var finals = PolymodFinalMacro.getAllFinals().get(imp.fullPath);
```
EliteMasterEric, line 325: I prefer the original implementation of this line as it leaves `finals` as a non-nullable value.

EliteMasterEric (changes requested):

PR #233, opened 2025-07-17
[FIX] Call arguments are now only evaluted once

### LINKED ISSUE #231: [BUG] [experimental] The most deranged context loss error of all time
Minimal reproduction:

```haxe
  public function onUpdate(event:ScriptEvent):Void {
    super.onUpdate();

    if (FlxG.keys.justPressed.HOME) {
      trace(test1(event) && test2());
    }
  }

  public function test1(event:ScriptEvent) {
    return event.type == "UPDATE";
  }

  public function test2(event:ScriptEvent) {
    return true;
  }
```

This gives the following error:
```
Error while executing function test1()#20:
Tried to access "type", but it is not a valid field or method. Is the target object null?
```

Note the following:

- The call to `test2()` is missing an argument. If the argument is properly provided, the error does not occur, thus making this a low-priority fix.
- Note that the issue complains about `test1()`. This is despite the fact that it is called BEFORE test2.
- Calling `trace(test2())` then `trace(event)` prints `null`, suggesting the variable is being deleted when returning to the original scope. So why is `test1()` affected?
- This only has a problem on the `experimental` branch, `master` is fine.

lemz1:
## Linked Issues

Fixes #231 (Fixes only the function arguments being evaluated twice)

## Description

I forgot that `hscript.Interp.expr(ECall)` evaluates the function parameters as well, so now we handle ECall directly in the overriden function, instead of falling back to `super.expr()`;

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -663,8 +663,15 @@ class PolymodInterpEx extends Interp
 						if (name != null && _scriptEnumDescriptors.exists(name))
 						{
 							return new PolymodEnum(_scriptEnumDescriptors.get(name), f, args);
+						} 
+						else 
+						{
+							var obj = expr(e);
+							if( obj == null ) errorEx(EInvalidAccess(f));
+							return fcall(obj,f,args);
 						}
 					default:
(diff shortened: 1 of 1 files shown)


lemz1:
#234 Fixes the entirety of #231

Issue #238, opened 2025-07-27
[BUG] [experimental] Mod assets get included repeatedly in every Lime library


NotHyper-474:
Since FNF 0.7.2 there have been reports of some stuff taking way more time to load than compared to previous versions. This is due to various uses of `Assets.list` throughout the game, but the problem isn't the function itself but rather how Polymod is affecting it.

Simply debugging it shows, as the title points out, that mod assets are being included repeatedly in all libraries. The way it affects `Assets.list` is shown below.

https://github.com/user-attachments/assets/fae5bc06-f7b5-413c-8385-8499016543d6

NotHyper-474:
While trying to fix this I've discovered what was causing Polymod to seemingly "ignore" the provided type when doing `Assets.list` (someone forgot to use `stripAssetsPrefix`.......) and while at it found that all mod audio files are treated as `AUDIO_SOUND` which is problematic as if the former issue is fixed then any music asset from a mod is going to be treated as non-existent. This is already enough of a clusterfuck.

Hundrec:
Yay

Hundrec:
I'm gonna push for this one

Hundrec:
🍊👍

PR #240, opened 2025-07-31
[OPTIMIZATION] Optimize listing modded assets from libraries

KoloInDaCrib:
Closes https://github.com/larsiusprime/polymod/issues/238

Hyper pointed it out to me in a DM that around 8 months ago there has been made a commit that made every lime library list every asset from the game, regardless of if the library itself has the asset or not. This PR fixes it by implementing library checking in `PolymodAssetLibrary`.


EliteMasterEric:
Still need to test that this doesn't break anything but from the code review I did, this looks good.

PR #242, opened 2025-08-01
[FIX] Fix variable declaration cache not actually being used

NotHyper-474:
I think the diff speaks for itself


EliteMasterEric:
i-fired-and-i-missed-again

PR #243, opened 2025-08-01
[ENHANCEMENT] Improve properties support

NotHyper-474:
Improves support for properties so that they act more like Haxe ones.

* Accessing a property inside its get/set won't cause infinite recursion anymore.
* Properties with both a getter and a setter will act like they don't have a "physical field", so trying to access it inside them will yield an error, just like in Haxe. Adding `@:isVar` to the field will prevent this.
* Fixes postfix increment not acting like it should.


EliteMasterEric:
Works good, code makes sense.

Needs a merge conflict resolution though.

EliteMasterEric (approved):

PR #246, opened 2025-08-11
[FIX] Fix yet another case of context loss

NotHyper-474:
### Minimal repro
```haxe
function func1(event)
{
  func2(); // This function causes "event" to be deleted after `func1` is exited
  trace(event); 
}

function func2()
{
  //yadayada
}

public override function onUpdate(event:ScriptEvent):Void
{
  if (FlxG.keys.justPressed.HOME)
  {
    func1(event);
    trace(event); // With the variable now lost this line causes an EInvalidAccess("event") error
  }
}
```
This is because `previousValues` is now a static Map rather than part of the execution's local scope (which was apparently done as an optimization), which means all scripted functions share the same set of previous values. So every time a function is called inside another the previous context gets lost.
And the argument becomes a figment of our minds in this snippet. Since the previous value doesn't exist in `previousValues` anymore whatever existing variable with the same name ends up being removed.
https://github.com/larsiusprime/polymod/blob/0d2868e53d45f2cbad7f9255f617aafd98d33cf3/polymod/hscript/_internal/PolymodScriptClass.hx#L625-L635

I think though that a proper fix would be to revert to the previous code where it's a local variable, or maybe append the function name to the argument names so that existing variables aren't prone to issues.

### DIFF
--- a/polymod/hscript/_internal/PolymodScriptClass.hx
+++ b/polymod/hscript/_internal/PolymodScriptClass.hx
@@ -564,8 +564,6 @@ class PolymodScriptClass
 		}
 	}
 
-	static final previousValues:Map<String, Dynamic> = [];
-
 	@:privateAccess(hscript.Interp)
 	public function callFunction(fnName:String, args:Array<Dynamic> = null):Dynamic
 	{
@@ -575,7 +573,7 @@ class PolymodScriptClass
 		if (fn != null)
 		{
 			// previousValues is used to restore variables after they are shadowed in the local scope.
-			previousValues.clear();
+			var previousValues:Map<String, Dynamic> = [];
 			var i = 0;
 			for (a in fn.args)
 			{


EliteMasterEric:
> Perhaps it's best to make `previousValues` a local variable like it used to be.

Before we do that, make sure to check WHY previousValues was changed to a static variable previously, it may be that that change was made to fix a DIFFERENT bug and we'll have to restructure things entirely.

NotHyper-474:
It was changed to a static variable in commit https://github.com/larsiusprime/polymod/commit/6a3f3793c29ec2922630854ba5186f6f1d73d25a so apparently it was for an optimization. I tested out changing it back to a local variable and tried songs where scripts are used (like Stress, Blazin' for instance) and there didn't seem to be any apparent issues.

I also decided to try a different approach where `previousValues` is a instance variable, and the arguments are added to it with the function's name appended to the name (e.g `func1#event`), and it appears to work identically. The performance difference was negligible though.

PR #249, opened 2025-08-20
Fix ZIP assets being listed as lowercase when using case-insensitive config

NotHyper-474:
This was the result of an attempt at optimizing something that went real wrong. For instance trying to load QT-Rewired as a mod this caused all characters to fail to load since the mod made them all uppercase for some reason.


EliteMasterEric:
Wait, does case insensitive asset loading still work?

PR #255, opened 2025-09-23
[BUGFIX] Fix properties wiping local variables in local functions

NotHyper-474:
Back to fix some more issues with those darn properties. I did some overall refactor to how properties are handled (yes, I know I said I didn't want to use try-catching) and fixed an issue where doing anything with a property's getter/setter wipes local variables due to how `callFunction` works. The fix basically involves doing what `PolymodInterpEx#callThis` does.

I also did a minor change that could improve performance a bit methinks where every use of `findVar` will only look for vars in the cache (didn't do that in `PolymodAbstractScriptClass` from uncertainty)


NotHyper-474:
Superseded by #257

PR #257, opened 2025-09-23
[BUGFIX] Fix locals getting dropped after calling script's own `scriptCall`

KoloInDaCrib:
Basically fixes locals getting dropped when doing something like this
```
function onCreate(event:ScriptEvent)
{
    var i:Int = 0;
    scriptCall("americaYa");
    trace(i); // Normally this would cause a NOR
}

function americaYa()
{
    trace("hallo :D");
}
```

I wouldnt be surprised if this change fixes other instances of locals getting dropped since I modified PolymodScriptClass' `callFunction` function.


EliteMasterEric:
Code LGTM

PR #258, opened 2025-09-25
Fixes for HashLink

NotHyper-474:
Fixes some issues with HashLink: an error when trying to resolve abstracts, and a workaround for https://github.com/HaxeFoundation/hscript/issues/96 by using a macro that creates a wrapper class for `Std` and `Math`.

Edit: Now also fixes enum values not being evaluated properly

### DIFF
--- a/polymod/hscript/_internal/HLWrapperMacro.hx
+++ b/polymod/hscript/_internal/HLWrapperMacro.hx
@@ -1,0 +1,187 @@
+package polymod.hscript._internal;
+
+#if (!macro && hl)
+@:build(polymod.hscript._internal.HLWrapperMacro.buildWrapperClass())
+class HLMath extends Math {}
+
+@:build(polymod.hscript._internal.HLWrapperMacro.buildWrapperClass())
+@:haxe.warning("-WDeprecated")
+class HLStd extends Std
+{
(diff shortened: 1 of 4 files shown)


NotHyper-474:
Hmm I just realized that in hxcpp if you try to get a value from a enum that doesn't exist it'll just return null, meanwhile my last commit made it so that in HashLink it throws an error. Maybe it should be reverted.

EliteMasterEric:
Is this fully functional or still in testing?

NotHyper-474:
I guess it stills needs better parity with HXCPP (which I don't think will ever be 100% achievable)

NotHyper-474:
I think this is good enough to at least be able to run FNF in HashLink. I don't think there's a way to fix all of the parity issues, so don't expect cross-target compatibility of mods to work.

EliteMasterEric:
This PR is kind of in limbo since I can't get lime to build anything with hashlink right now 🤷 

I'm also not too much of a fan of HashLink anyway so I can't really be bothered to put a lot of effort into this.

NotHyper-474:
That's understandable. But if you'd like to test it sometime, there's a list of all the necessary changes to run in the PRs listed in https://github.com/FunkinCrew/Funkin/pull/6135, including a fix for the Lime build issue.

PR #259, opened 2025-10-06
[OPTIMIZATION] Optimize checking for files

KoloInDaCrib:
File checking is done in a manner where `fileSystem.exists` is called for every mod installed, even if they contain nothing but the polymod meta file. This is really unoptimized and rears its ugly head especially with too many mods installed, most noticeable in the Freeplay menu upon changing difficulties where each individual pixel icon uses 3 `Assets.exists` calls (each one doing the polymod file checking logic.)

This PR replaces that with a pre-filled array of assets that are loaded with mods, The array is cleared and re-filled upon further Polymod reinitializations.

> [!NOTE]
> I do not have a Linux device and I am not sure if this works on Linux considering a lot of checks regarding its file capitalizations were written.


KoloInDaCrib:
Superceded by #260

PR #261, opened 2025-10-25
[ENHANCEMENT] Display an error when redefining a variable

KoloInDaCrib:
If one were to redefine a variable from the super class inside a scripted class, it could potentially cause a crash. This PR displays a warning whenever that happens so that modders aren't left confused for seemingly no reason.
This only applies to non-static variables - functions are left intact (since otherwise it'd be a breaking change lmao)

image


EliteMasterEric:
We definitely want to make sure users aren't defining variables that override existing ones, since there is no case where you'd want to do it and it can cause unintended behavior if done on accident.

The default behavior for functions should still be to assume the user wants to override (as long as that function hasn't already been declared in the same script, but I think I fixed that issue a bit ago).

PR #263, opened 2025-11-22
[FIX] Error if HScript isn't installed

CrusherNotDrip:
If hscript isn't installed, it usually won't register script classes but now it just prevents you from compiling all together and this pr fixes that by returning an empty array if hscript isn't installed.


TechnikTil:
holy run-on sentence

CrusherNotDrip:
> holy run-on sentence

address the scream in the background

TechnikTil:
how am i supposed to know

EliteMasterEric (approved):

PR #264, opened 2025-11-29
[BUGFIX] Stop static class declaration from getting lost in local functions

NotHyper-474:
Whenever you have a local function that was created by a static function and it is called by something that isn't a scripted class, the `_classDeclOverride` would be null, so any imports and stuff would be missing from the context.

Here's an example script where this happens
```haxe
import funkin.modding.base.Module;
import flixel.util.FlxTimer;
import flixel.FlxG;

class Repro_264 extends Module
{
	function new()
	{
		super('PR_264');
	}

	static function foo()
	{
		FlxTimer.wait(0.1, () -> {
			// Error while executing function ???.anonymous()#17
			// Tried to access "FlxTimer", an unknown variable or identifier.
			new FlxTimer();
		});
	}

	override function onUpdate(e)
	{
		if (FlxG.keys.justPressed.HOME) foo();
	}
}
```

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -549,6 +549,7 @@ class PolymodInterpEx extends Interp
 				var capturedLocals = duplicate(locals);
 				var me = this;
 				var hasOpt = false, minParams = 0;
+				var capturedClassDecl = _classDeclOverride;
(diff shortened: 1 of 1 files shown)


NotHyper-474:
Just realized this is also done by https://github.com/larsiusprime/polymod/pull/209 and even more

PR #266, opened 2025-11-30
Use `superHasField` instead of static field cache in `resolveField`

NotHyper-474:
Instead of using the static cache introduced by #221 we can simply use `superHasField` which basically does the same thing.

~We don't explicitly check for the getter or setter field anymore since `Reflect`'s property functions already check for those.~ Turns out that IS needed.

We also remove the concatenation with `Reflect.fields` when caching since that feels redundant and just creates duplicate entries in the array.

### DIFF
--- a/polymod/hscript/_internal/PolymodAbstractScriptClass.hx
+++ b/polymod/hscript/_internal/PolymodAbstractScriptClass.hx
@@ -6,8 +6,6 @@ import haxe.ds.ObjectMap;
 @:access(polymod.hscript._internal.PolymodScriptClass)
 abstract PolymodAbstractScriptClass(PolymodScriptClass) from PolymodScriptClass
 {
-  static final fieldsCache:ObjectMap<Dynamic, Array<String>> = new ObjectMap();
-
   private function resolveField(name:String):Dynamic
   {
     switch (name)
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
We need to make sure this doesn't break anything before we merge it.

NotHyper-474:
> We need to make sure this doesn't break anything before we merge it.

I've changed it so it keeps fields obtained from `Reflect.fields` and just removes any duplicates which hopefully is less prone to breaking things.
Edit: Turns out what broke things was actually something else but I got that covered.

NotHyper-474:
Nevermind this breaks some stuff anyway lol

polymod/hscript/_internal/PolymodScriptClass.hx
```diff
@@ -434,7 +434,9 @@ class PolymodScriptClass
 		// Reflect.hasField(this, name) is REALLY expensive so we use a cache.
 		if (__superClassFieldList == null)
 		{
-			__superClassFieldList = Reflect.fields(superClass).concat(Type.getInstanceFields(Type.getClass(superClass)));
+			var instanceFields = Type.getInstanceFields(Type.getClass(superClass));
+			__superClassFieldList = Reflect.fields(superClass);
+			__superClassFieldList = __superClassFieldList.concat(instanceFields.filter(f -> !__superClassFieldList.contains(f)));
```
EliteMasterEric: Ah, this is preventing duplicates in __superClassFieldList, makes sense.

NotHyper-474:
Honestly, I don't see the point of doing this anymore

PR #268, opened 2025-12-14
A possible "fix" for the code completion/Intellisense in Funkin'

NotHyper-474:
The gist of it is that we prevent `HScriptedClassMacro` of ever running on the language server because when it does so code completion will stop working. I have no idea what in the macro causes this issue; I feel like it could be a problem with Haxe itself. Unfortunately while it works, the code completion will be painfully slow thanks to HaxeUI macros.

Note that `#if display` simply doesn't work (this is mentioned in the comment I left in the code), the only thing that did was checking `Context.getDisplayMode()`.

### DIFF
--- a/polymod/hscript/_internal/HScriptedClassMacro.hx
+++ b/polymod/hscript/_internal/HScriptedClassMacro.hx
@@ -19,6 +19,11 @@ class HScriptedClassMacro
 	 */
 	public static macro function build():Array<Field>
 	{
+		// This macro seems to cause issues in the language server in macro-heavy projects,
+		// so we prevent it to run on display (for some reason the define doesn't really work so we have to check at runtime)
+		// TODO: Remove this when whatever causes those issues is resolved.
+		if (Context.getDisplayMode() != None) return null;
+
 		var cls:haxe.macro.Type.ClassType = Context.getLocalClass().get();
 		var initialFields:Array<Field> = Context.getBuildFields();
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
This worked but only for one or two inspects, after that I start getting this error (I see it in VSCode -> Output -> Haxe):

> [Error - 1:02:23 PM] Request textDocument/hover failed.
>  Message: Uncaught exception Stack overflow
>  Code: -32603 

If I had a better way of getting stack traces I could give more info.

NotHyper-474:
Weird, both completion and hover hints work for me no matter how much I trigger them. There's a way of getting the stack trace in the form of a very verbose JSON by changing the display server arguments to the following, but I don't think it'd be of any use.
 ```json
 "haxe.displayServer": {
    "arguments": ["-v"],
    "print": {
      "completion": true,
      "reusing": false
    }
  }
```

I'll rebase this on top of the latest commit and mention this PR at https://github.com/FunkinCrew/Funkin/issues/4147 to see if the folks over there get the same issue.

NotHyper-474:
Ok, actually I just ran into that issue myself. Apparently it's the same error you were running into in https://github.com/FunkinCrew/funkinBounties/issues/4 which came back because the `haxe.enableCompilationServer` got enabled. Honestly I think the only way to avoid it would be to disable it again.

T5mpler:
this works you are actually my dad dude thank you I've hated having to deal with this for the past 4 months

PR #270, opened 2025-12-24
Fix an issue with blacklisted fields maps

NotHyper-474:
Fixes whatever this bullshit is
image

This is causing objects to be forcibly converted into strings when passed into this check, which can be an issue because destroyed Flixel objects can try to be passed into it and cause crashes.

https://github.com/larsiusprime/polymod/blob/f56cbf3edc594766d51c413cec7fc1008ca6c886/polymod/hscript/_internal/PolymodInterpEx.hx#L1131

### DIFF
--- a/polymod/Polymod.hx
+++ b/polymod/Polymod.hx
@@ -872,7 +872,7 @@ class Polymod
 	public static function blacklistInstanceFields(parentClass:Class<Dynamic>, fields:Array<String>):Void
 	{
 		#if hscript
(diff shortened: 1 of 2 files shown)


Hundrec (approved):
It fixes the Stress crash!

AbnormalPoof (approved):
Let's goooo!!! It fixes it!

NotHyper-474:
whattheshit

PR #271, opened 2025-12-25
[FIX] Fix class name parsing for some objects

NotHyper-474:
> [!NOTE]
>
> This PR now fixes some class types getting parsed to "Object" as if they were anonymous structures. You can still check out the original description below.

## Original Description

https://github.com/larsiusprime/polymod/blob/7f82ddb3e11e33ad0dd42e0c3e1417edc90e38f9/polymod/hscript/_internal/PolymodInterpEx.hx#L1124-L1125

This snippet will convert any object to a string; some anon structures can contain string maps in them which will allocate memory when those are converted to strings when getting or setting a value. One a lot of mods seem to reference is `FlxG.save.data` which can contain a lot of string maps, causing very frequent memory allocations.

https://github.com/user-attachments/assets/a53edb9b-effc-4b90-ad09-b97d87129c10

Since a TObject can be either a class type or a struct I was compelled to either:
 * Use `Type.getClassName` (which is apparently slow even though all it does is access a field behind the scenes? Not saying it is, I just heard it from someone)
 * Take advantage of the fact Polymod has to resolve classes from imports and when initializing objects, and so store these classes in a map along with their names. (Shout out to Kolo for the idea)

I decided to go with the second since `getClassName` has so much hatred. I haven't had the chance to test it thoroughly but testing the blacklist of `FlxSave.resolveFlixelClasses` (static) and `FlxSave.data` (instance) and both worked, and the memory leak disappeared.

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -1256,7 +1256,7 @@ class PolymodInterpEx extends Interp
   {
     if (o == null) errorEx(ENullObjectReference(f));
 
-    var oCls:String = Util.getTypeName(Type.typeof(o));
+    var oCls:String = Util.getTypeNameOf(o);
 
     // Check if the field is a blacklisted static field.
     if (PolymodScriptClass.blacklistedStaticFields.exists(o) && PolymodScriptClass.blacklistedStaticFields.get(o).contains(f))
@@ -1266,7 +1266,7 @@ class PolymodInterpEx extends Interp
     }
 
     // If not, check if it is a blacklisted instance field.
-    if (oCls.length > 0)
+    if (oCls.length > 0 && oCls != 'Object')
     {
       if (PolymodScriptClass.blacklistedInstanceFields.exists(oCls) && PolymodScriptClass.blacklistedInstanceFields.get(oCls).contains(f))
       {
@@ -1337,7 +1337,7 @@ class PolymodInterpEx extends Interp
       // return result;
     }
 
-    var abstractKey:String = Type.getClassName(o) + '.' + f;
+    var abstractKey:String = '$oCls.$f';
     if (PolymodScriptClass.abstractClassStatics.exists(abstractKey))
     {
       return Reflect.getProperty(PolymodScriptClass.abstractClassStatics[abstractKey], abstractKey.replace('.', '_'));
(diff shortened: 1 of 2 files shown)


polymod/fs/SysFileSystem.hx
```diff
@@ -108,7 +108,6 @@ class SysFileSystem implements IFileSystem
 	public function getMetadata(modId:String)
 	{
 		var modPath = Util.pathJoin(modRoot, modId);
-		var test = readDirectory(modRoot);
```
EliteMasterEric: Is this change unrelated? Should be its own PR.

polymod/hscript/_internal/PolymodInterpEx.hx
```diff
@@ -27,6 +28,13 @@ class PolymodInterpEx extends Interp
 
 	var _propTrack:Map<String,Bool> = [];
 
+	static final _knownClassTypes:haxe.ds.ObjectMap<Dynamic, String> = new haxe.ds.ObjectMap();
```
EliteMasterEric: This should probably be in Util with the function that gets the type name.

EliteMasterEric (changes requested):
Since I fixed TObject to return String, the only thing this PR does is provide some caching for certain known types, right? What do you mean by TObject sometimes being a class type?

polymod/fs/SysFileSystem.hx
NotHyper-474: Yeah it was just an unnecessary unused variable declaration, which Haxe won't optimize out, that I thought would be fine to include here since it's just a one line change.

NotHyper-474:
> Since I fixed TObject to return String, the only thing this PR does is provide some caching for certain known types, right? What do you mean by TObject sometimes being a class type?

Yeah, I guess. All the cache is for is to store the names of classes returned by `Type.resolveClass` which will always be of the type `TObject` for some reason (perhaps it's a Haxe bug?), and you can see that's the case [here](https://try.haxe.org/#07310ed0).
And all this is to avoid having to use `Type.getClassName`, unless you would actually be fine with that.

EliteMasterEric:
I just looked at the code internally, and all `Type.getClassName` does on C++ is access a string value on the class object.

Not sure what any performance issues would be with that, so if we could use that instead and remove the extra code that would be nice.
image

NotHyper-474:
> I just looked at the code internally, and all `Type.getClassName` does on C++ is access a string value on the class object.

Perfect, that was exactly what I was confused about. I thought the whole ordeal of using `Type.typeof` was to avoid using `Type.getClassName` but I also don't think it has any impact on performance at all. I'll be making the suggested changes.

EliteMasterEric (approved):

NotHyper-474:
oh, whoops lol

PR #272, opened 2025-12-29
Remove unused call to `readDirectory`

NotHyper-474:
Removes an unused local variable that isn't optimized out in `SysZipFileSystem`.
Made as per request of https://github.com/larsiusprime/polymod/pull/271#discussion_r2651403344


NotHyper-474:
image

Funny how only the variable definition gets optimized by the static analyzer, and not the function call (speed trying not to laugh)

EliteMasterEric:
> Funny how only the variable definition gets optimized by the static analyzer, and not the function call (speed trying not to laugh)

Probably because the compiler thinks there might be side effects.

Will double check this before merging.

AbnormalPoof (approved):
lol

PR #273, opened 2026-01-09
Make text asset caching opt-in

NotHyper-474:
This PR aims to fix this quite niche issue (https://github.com/FunkinCrew/Funkin/issues/6675) which was introduced by the assets caching PR.

To sum it up, if you have a mod that constantly needs to modify a text file and later on you have to get the contents you've written, you'll just get the same thing instead of what was written to it due to the contents getting cached.

Since I don't think retrieving text files is something that happens a lot, or at least not in the base game, I think it's fine to disable this caching by default (although my actual reasoning for this is because I know that even if this is merged as an opt-out someone will forget to disable it)


PurSnake:
goodies

spark-the-dragon:
I love you thank you for this

PR #274, opened 2026-01-11
[FIX] Call super class constructor arguments properly

NotHyper-474:
Currently, the way `super` calls the super class' constructor in Polymod is a bit jank, because not only any null arguments get skipped but also anything past 4 arguments.
We can't just call `createSuperClass` directly because it has an array as the argument, but we can use `Reflect.makeVarArgs` to turn it into a function that, when called by `Reflect.callMethod`, will correctly accept the arguments as an array.

One issue this solves is this case with a scriptable note kind, which is currently borked.
```hx
package test.notekind;

import funkin.play.notes.notekind.ScriptedNoteKind;

class CustomAnimNoteKind extends ScriptedNoteKind
{
  final DISABLE_ANIMATIONS:Bool = true;

  public function new()
  {
    super("somekind", "Are we broken?", null, [], DISABLE_ANIMATIONS);
    trace(this.noteKind);
    trace(this.description);
    trace(this.noteStyleId);
    trace(this.params);
    trace(this.noanim);
    trace(this.suffix);
  }
}
```

Despite the arguments being correct, the fields are not.
image

With this fix, that doesn't happen anymore.
image

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -1303,7 +1303,7 @@ class PolymodInterpEx extends Interp
 			}
 			else if (_proxy.superClass == null)
 			{
-				return _proxy.superConstructor;
+				return Reflect.makeVarArgs(_proxy.createSuperClass);
 			}
 			else
 			{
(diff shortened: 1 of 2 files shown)


Starexify (approved):

EliteMasterEric (approved):
Great PR! The explanation of the fix and the minimal repro greatly improve the review process.

PR #278, opened 2026-01-13
Simplify multi-argument function calling

NotHyper-474:
Instead of a giant switch case just to handle a bunch of different amounts of arguments (which even is limited) we can just use `Reflect.makeVarArgs` which is simple and has no limits for arguments (except on some targets).


Starexify:
Tested this and it works perfectly:
image
<sub>it also seems that there are no performance issues with this approach<sub>

With current polymod this happens:
image

Starexify (approved):

EliteMasterEric (approved):
Works perfectly yayyyyy

PR #279, opened 2026-01-16
[FIX] Fix default argument values not being used by functions/lambdas when the passed values are missing or null.

Starexify:
This PR fixes the way default argument values are handled when called a function/lambda.

Currently setting a default argument in functions won't use the argument when the passed argument is `null` or just missing. This was caused because of the `if` statement not checking correctly for the argument passed, and instead just checking if arguments exist at all or if the argument length matches, so it never reached the `a.value` part to set the default function value.

For lambdas, the minimum argument requirement shouldn't count default values if they exist, so this PR also fixes that. Also default values are applied if the value is `null` or just missing.

### With this PR:

image

image

### Without this PR:

image

image

### Haxe behavior
Even in Try haxe it's the same behavior:

Functions: https://try.haxe.org/#F5c4e9b2
Lambdas: https://try.haxe.org/#96735bBF

image

image

### DIFF
--- a/polymod/hscript/_internal/Parser.hx
+++ b/polymod/hscript/_internal/Parser.hx
@@ -398,6 +398,27 @@ class Parser
     return parseExprNext(mk(EObject(fl), p1));
   }
 
+  /**
+   * Transform an Expr to an Argument for lambda functions
+   *
+   * @param e The expression
+   * @return The function argument
+   */
+  function exprToArg(e:Expr):Argument {
+    return switch (expr(e)) {
+      case EIdent(v):
+        {name: v, t: null, opt: false, value: null};
+      case ECheckType({e: EIdent(v)}, t):
+        {name: v, t: t, opt: false, value: null};
+      case EBinop("=", {e: EIdent(v)}, def):
+        {name: v, t: null, opt: false, value: def};
+      case EBinop("=", {e: ECheckType({e: EIdent(v)}, t)}, def):
+        {name: v, t: t, opt: false, value: def};
+      default:
+        null;
+    }
+  }
+
   function parseExpr()
   {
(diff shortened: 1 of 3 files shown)


EliteMasterEric:
Shouldn't `test()` throw an error (because the minimum number of arguments was not passed)?

Starexify:
> Shouldn't `test()` throw an error (because the minimum number of arguments was not passed)?

Yes indeed, but I dont remember changing the logic for that so it's original behavior, this only focused on the default arguments issue only 🤔

EliteMasterEric:
We should probably make it so the minimal argument count matters, and throw an error if the arguments aren't provided.

Starexify:
> We should probably make it so the minimal argument count matters, and throw an error if the arguments aren't provided.

Sure, will update the PR with that feature for functions aswell, lambdas did that and I was wondering why they do but functions dont, but true they should have a consistency to real haxe 👍

Starexify:
> We should probably make it so the minimal argument count matters, and throw an error if the arguments aren't provided.

So:

I fixed the argument count check for normal functions with a better approach to check for the min count by finding the rightmost argument without a default value or optional operator.
And I might need to implement it in lambdas aswell.

And some other issues with lambdas and defaults not working in the first parameter that I'll try to fix

Starexify:
Alright now the optional/default arguments should work now for lambdas and functions.
And the minimum parameter count is correctly checked for both aswell.

EliteMasterEric:
This can be done in a future PR but `ECustom('Invalid number of parameters')` should definitely be replaced with a custom value like `EInvalidParamCount()`

EliteMasterEric:
No other critiques LGTM

PR #280, opened 2026-01-16
Allow scripted classes to not extend anything

NotHyper-474:
Fixes a few instances where classes that don't extend another class aren't properly accounted for.

### DIFF
--- a/polymod/hscript/_internal/HScriptedClassMacro.hx
+++ b/polymod/hscript/_internal/HScriptedClassMacro.hx
@@ -906,26 +906,37 @@ class HScriptedClassMacro
 						ret: func_ret,
 						expr: macro
 						{
-							if (_asc != null && _asc.hasScriptFunction($v{funcName}))
+							if (_asc != null)
 							{
-								// trace('ASC: Calling $fieldName() in macro-generated function...');
-								$
+								if (_asc.hasScriptFunction($v{funcName}))
 								{
-									doesReturnVoid
-										? (macro _asc.callFunction($v{funcName}, [$a{func_callArgs}]))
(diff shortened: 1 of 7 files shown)


Starexify (approved):

EliteMasterEric (changes requested):
A couple things to clean up:

- The error when calling super() on a parentless class constructor should probably be cleaned up, it's displaying the same line number twice. It looks like the `ECustom` is getting wrapped in an `EScriptThrow`?

image

- Calling `toString()` on a parentless object throws an error, it should call `Std.string(v:Dynamic)` if the function isn't overridden.

Image

NotHyper-474:
> * The error when calling super() on a parentless class constructor should probably be cleaned up, it's displaying the same line number twice. It looks like the `ECustom` is getting wrapped in an `EScriptThrow`?

Fixed. Turns out `PolymodInterpEx.call` was wrapping any error passed to it in a `EScriptThrow` without considering whether it's a Polymod or a Hscript error. I've made it use a new `EInvalidSuper` error though, I didn't like how it called it an "unknown error".

> * Calling `toString()` on a parentless object throws an error, it should call `Std.string(v:Dynamic)` if the function isn't overridden. 

Should be fixed as well, I suppose it returning `PolymodScriptClass<ClassName>` as well is fine?

EliteMasterEric:
Code review looks good, just need to do some testing after I fix up a couple changes I'm working on locally.

NotHyper-474:
Added an extra commit with a few fixes for the class extension system that wouldn't work as a separate PR, you can check its description for details.

EliteMasterEric:
lmao i randomly looked at this commit and it was updated 7 minutes ago

time to review

EliteMasterEric (approved):
Works good

PR #281, opened 2026-01-16
Fix local variables persisting in static functions

### LINKED ISSUE #277: [BUG] Calling a static function after another static function can drops its argument values
## Description

When I try to use this example script:

```haxe
import funkin.modding.events.ScriptEvent;
import funkin.modding.module.ScriptedModule;

class ShenaniganOne extends ScriptedModule {
    /**
     * Calling a setter and static function crashes the game
     *
     * If `value` is changed to a diff name for the variable or argument it will work again
     */
    static var _instance:ShenaniganOne;
    public static var instance(get, set):ShenaniganOne;
    static function set_instance(value:ShenaniganOne):ShenaniganOne { return _instance = value; }
    static function get_instance():ShenaniganOne { return _instance; }

    public static function testFunc() {
        var value;
    }

    override public function onCreate(event:ScriptEvent) {
        testFunc();
        ShenaniganOne.instance = this;
    }
}
```

polymod crashes with the following log:
```
Null Object Reference
 in polymod/hscript/_internal/PolymodScriptClass.hx#640
 in polymod/hscript/_internal/HScriptedClassMacro.hx#917
 in funkin/modding/events/ScriptEventDispatcher.hx#34
 in funkin/modding/module/ModuleHandler.hx#157
 in funkin/InitState.hx#294
 in funkin/InitState.hx#72
 in flixel/FlxGame.hx#588
 in flixel/FlxGame.hx#340
```

If I changed the name of `var value;` or the argument `(value:Test)` from the setter to something else, the crash doesn't happen.

The issue seems to be related to this: [FunkinCrew/Funkin#5765](https://github.com/FunkinCrew/Funkin/issues/5765)

NotHyper-474:
Fixes #277

This addresses a few things I've noticed:

* Checking `locals` in `resolve` is practically useless, since `locals` will have already been checked before hscript's `expr` calls it. I've debugged, and the only instance where `locals` were checked here was when they persisted due to a bug.
* Reverts the change of `executeEx` to `exprReturn`, seems like the bug that was fixed by it was subsequently resolved by a different change. This solves local variables persisting in static functions.
* Copies the fix from #257 to apply it to static functions; solves static properties dropping local variables.

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -1489,11 +1489,6 @@ class PolymodInterpEx extends Interp
       return null;
     }
 
-    if (locals.exists(id))
-    {
-      // NOTE: id may exist but be null
-      return locals.get(id).r;
-    }
     if (variables.exists(id))
     {
       // NOTE: id may exist but be null
@@ -1681,10 +1676,11 @@ class PolymodInterpEx extends Interp
 
       this._classDeclOverride = cls;
 
+      var localsCopy:Map<String, {r:Dynamic, ?isfinal:Null<Bool>}> = this.locals.copy();
       var result:Dynamic = null;
       try
       {
-        result = this.exprReturn(fn.expr);
+        result = this.executeEx(fn.expr);
       }
       catch (err:PolymodExprEx.ErrorEx)
       {
@@ -1716,6 +1712,7 @@ class PolymodInterpEx extends Interp
         }
       }
       this._classDeclOverride = previousClassDecl;
+      this.locals = localsCopy;
 
       return result;
     }


Starexify (approved):

NotHyper-474:
Also, the bug I'm talking about in the second point used to affect this repro script.
```haxe
import funkin.modding.base.Object;

class StaticFunctionRepro extends Object
{
  public static function init():Void
  {
    var save = {version: "1.0"};
    var result = StaticFunctionRepro.checkSaveData(save);
    trace(result);
  }

  public static function checkSaveData(saveData:Dynamic):Dynamic
  {
    if (saveData.version != "2.0")
    {
      trace('Save data is outdated! Migrating...');
      saveData = migrateSaveData(saveData);
    }

    trace('Save data is good!');

    var updatedSaveData:Dynamic = getDefaultSaveData();
    return updatedSaveData;
  }

  public static function migrateSaveData(saveData:Dynamic):Dynamic
  {
    saveData.version = "2.0";
    return saveData;
  }

  public static function getDefaultSaveData():Dynamic
  {
    return {version: "2.0", other: true};
  }
}
```
After calling `init`, the local variables would act pretty weirdly, errors would be thrown as if they didn't exist. But now, even when replacing `exprReturn` with `executeEx` in the interp, which should've re-introduced the issue, it doesn't happen anymore.

EliteMasterEric:
The original PR has a repro script that crashes without this PR and works fine with it, so I'd say this is good to go

PR #282, opened 2026-01-16
[FIX] Allow functions of `using` classes to be used in static contexts

NotHyper-474:
Until now these were only available to non-static functions. Currently it's not cached (but so are other static things)
The diff is a bit big because I decided to give stuff some refactor


Starexify (approved):

AbnormalPoof (approved):
Despite the scary diff, it works very nicely and doesn't break anything! Tested with both instance and static functions.

PR #284, opened 2026-01-18
[ENHANCEMENT] Allow interfaces to be parsed and no longer throw errors when they exist in scripts.

Starexify:
This PR allows interfaces to be parsed and not just throw errors, which would make the whole script no longer work, even when working classes are inside. 

For now this is just allowing people to add interfaces without getting errors if they have them in their code structure alongside classes/enums etc.

**Now**

image

**Before this would be shown**

image

Since ~~hscript~~polymod would return `unexpected(TId(ident));`

### DIFF
--- a/polymod/hscript/_internal/Expr.hx
+++ b/polymod/hscript/_internal/Expr.hx
@@ -147,6 +147,7 @@ enum ModuleDecl
   DClass(c:ClassDecl);
   DTypedef(c:TypeDecl);
   DEnum(e:EnumDecl);
+  DInterface(e:InterfaceDecl);
 }
 
 typedef ModuleType =
@@ -188,6 +189,14 @@ typedef TypeDecl =
 {
   > ModuleType,
   var t:CType;
+}
+
+typedef InterfaceDecl =
+{
+  > ModuleType,
+  var extend:Array<CType>;
+  var fields:Array<FieldDecl>;
+  var isExtern:Bool;
 }
 
 typedef FieldDecl =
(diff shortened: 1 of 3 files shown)


Starexify:
Copy paste from the hscript PR:
Also I tried in tryhaxe this snippet which shows how externs behave, `private`, `static`, `final` works, but not `override`/`inline`. So that's why I didnt allow the `inline` keyword if someone is wondering
```haxe
class Test {
	static function main() {
		var test = new TestClass();
		trace(TestClass.id);
		trace(test.nr);
		trace(test.test());
	}
}

class TestClass implements TestInterface {
	var name:String = "class";

	public static final id:Int = 1;

	public var nr:Int = 90;

	public function new() {}

	public function test():String {
		return "TEST";
	}
}

extern interface TestInterface extends TestInterfaceTwo {
	private var name:String;
	static final id:Int;
	var nr:Int;
}

interface TestInterfaceTwo {
	function test():String;
}
```

EliteMasterEric:
To clarify, this does not mandate that the scripted class implements the interface, right? It just parses the interface and stores the data for later?

Starexify:
> To clarify, this does not mandate that the scripted class implements the interface, right? It just parses the interface and stores the data for later?

Yes, this only parses the interfaces, nothing else. It doesnt handle any of the interface logic.
But they could be implement by classes before this PR if they were in separate files. This just makes them not throw errors and breaking the script file if theyre in the same file as actual script classes, and allows parsing if they are ever gonna be needed in the future.

EliteMasterEric:
Tested with the repro script and it does fix the issue, thanks for your hard work

PR #285, opened 2026-01-19
[FEATURE] Allow scripted classes to extend other scripted classes

KoloInDaCrib:
This PR does what it says in the title.

image
image

The extends have been moved to `validateImports` to check for any scripted classes registered after the scripts have been loaded. To add to this, I've also added a feature to automatically import classes with the same or the parent package, since it felt weird writing `import Adversary;` in the same file as the class itself.

Like expected, `super` points to the parent script class. Using `this` also works on scripted fields`.


VirtuGuy:
This is peak.

EliteMasterEric (approved):
This one worked well for me with your example script.

PR #286, opened 2026-01-21
[Breaking Change] Rename `init` and make it not use the super constructor arguments

### LINKED ISSUE #241: [ENHANCEMENT] Rename `init` to `scriptInit` for creating scripted classes
Currently, to init a scripted class, you have to run `[ScriptedClass].init([name], [params]);`. However, due to the name of this function and due to the fact that Haxe doesn't support having an instance and a static function with the same name, several classes that already have an `init` function of their own such as `FlxSound` cannot be made scriptable.

My proposed solution would be to rename the initialization function to `scriptInit` and to implement some sort of backwards compatibility to `PolymodInterpEx`. That way, not only is the problem of some classes not being possible to script avoided, but it would also follow the naming convention of other `HScriptClassMacro` methods (`scriptGet`, `scriptSet`, `scriptCall`).

NotHyper-474:
Resolves #241 

`HScriptedClass.init` is pretty odd because it's used throughout FNF as the equivalent of `new` for script classes, but it accepts the super class's constructor arguments as its own, and then initializes an instance of the script class using those arguments, however a script class is actually independent from the super class in this regard and can have any arguments it wants, which may not be the same the super class has. When scripts use it, it causes issues in some targets, where some of the arguments either get converted incorrectly, become null or, on HashLink, throws an error about the incompatible type.

One instance of that in FNF is `phillyTrain.hxc`, which instantiates an instance of `BuildingEffectShader` with `ScriptedFlxRuntimeShader.init('BuildingEffectShader', 1.0)`, which calls an `init` of arguments (Null\<String>, Null\<String>, Null\<String>), it works fine on C++ since there is no type-checking, but on HL it throws an error about being unable to cast a float to a string. This is not a mistake on the script's end, since it does accept a float as its argument.

So I've made it use [haxe.Rest](https://api.haxe.org/haxe/Rest.html) for the arguments, however since it doesn't work well with Reflect I had to add some backwards compatibility handling for scripts using it.

The registries and other parts of Haxe code that call `init` seem to work well with these changes somehow, but I recommend caution regarding any breaking changes this could introduce.

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -1256,6 +1256,14 @@ class PolymodInterpEx extends Interp
   {
     if (o == null) errorEx(ENullObjectReference(f));
 
+    // Backwards compatibility for scripts using HScriptedClass.init
+    // Std.isOfType(o, HScriptedClass) only works with class instances
+    // so we look for a specific field to double check the type
+    if ((f == 'init' || f == 'scriptInit') && o._isHScriptedClass)
+    {
+      return Reflect.makeVarArgs((args:Array<Dynamic>) -> return o.scriptInit(args[0], args.slice(1)));
+    }
+
     var oCls:String = Util.getTypeNameOf(o);
 
     // Check if the field is a blacklisted static field.
--- a/polymod/hscript/HScriptedClass.hx
+++ b/polymod/hscript/HScriptedClass.hx
@@ -22,7 +22,7 @@ interface HScriptedClass
    * @param constArgs The arguments to pass to the constructor.
    * @return The instance of the scripted class.
    */
-  // public static function init(clsName:String, constArgs:Array<Dynamic>):ThisType;
+  // public static function scriptInit(clsName:String, ...constArgs:Dynamic):ThisType;
   /**
    * Call a custom static function on a scripted class, by the given name, with the given arguments.
    *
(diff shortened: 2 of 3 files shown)


polymod/hscript/_internal/PolymodInterpEx.hx
```diff
@@ -1184,6 +1184,14 @@ class PolymodInterpEx extends Interp
 		if (o == null)
 			errorEx(ENullObjectReference(f));
 
+		// Backwards compatibility for HScriptedClass.init since Haxe.Rest doesn't work with reflection.
+		// Std.isOfType(o, HScriptedClass) doesn't work if the type was retrieved from Type.resolveClass
+		// so to double check, we see if the function scriptStaticGet exists
+		if (f == 'init' && o.scriptStaticGet != null)
```
EliteMasterEric: My preferred method of doing this would be to add another field which is explicitly for this, just in case.

So something like `o._isHScriptedClass == true`.

EliteMasterEric (changes requested):
Aside from the change I linked, since you're making a potentially breaking change here and adding backwards compatibility for it anyway, it would be nice to implement #241 and rename `init` to `scriptInit` for consistency with the other generated methods.

NotHyper-474:
All done, though now that the function has received a new name it surely is a breaking change for existing Haxe code that used `init` now.

EliteMasterEric:
I want to test and merge this, it'll take a PR on the Funkin' side though.

Notably, internally I already made a PR to optimize out all the `init`, `scriptGet`, and `scriptSet` calls in scripts that access other scripts (you can just use the constructor and property accesses like a normal class). I'd just need to update the source code.

PR #287, opened 2026-01-24
Move the FunkinCrew/HScript fork to be built into Polymod

EliteMasterEric:
No longer will the library depend on a specific fork of a haxelib to work. This also means improvements to Polymod no longer require pull requests to a separate repo.

This (along with fixing up static fields in scripted classes, and fixing up abstract access) is one of my goals before finally packaging up and updating the Haxelib.

In the future, I'd like to eventually combine the code from stuff like `PolymodInterpEx` into the base classes from HScript to make it a concise re-implementation of HScript without the weird chain of overridden methods it has currently.

### DIFF
--- a/polymod/Polymod.hx
+++ b/polymod/Polymod.hx
@@ -8,9 +8,7 @@ import polymod.backends.PolymodAssets;
 import polymod.format.JsonHelp;
 import polymod.format.ParseRules;
 import polymod.fs.PolymodFileSystem;
-#if hscript
 import polymod.hscript._internal.PolymodScriptClass;
-#end
 import polymod.util.DependencyUtil;
 import polymod.util.VersionUtil;
 import thx.semver.Version;
@@ -327,7 +325,6 @@ class Polymod
 		prevParams = params;
 
 		// Do scripted class initialization now that the assetLibrary is loaded.
-		#if hscript
 		if (params.useScriptedClasses)
 		{
 			Polymod.notice(PolymodErrorCode.SCRIPT_PARSING, 'Parsing script classes...');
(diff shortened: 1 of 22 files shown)


EliteMasterEric:
I want to get this one merged quickly, so that I don't have to redo it due to PRs on this repository or the HScript one (thanks for the hard work @starexify!) but I want to make sure it doesn't break anything, so review from a bunch of the contributors here would be great.

Starexify:
Personally I'd merge polymod and hscript , even class-wise, so we dont have multiple classes and have to extend to modify and add functionality to the logic like Interp and PolymodInterpEx

Edit: or maybe leave it like this for now and then have another PR to merge the classes best to not complicate things

Starexify (approved):
Checked the code diff and tested with my mods on a funkin build without hscript and only this polymod PR and should work the same as before without issues. 👍

EliteMasterEric:
> Personally I'd merge polymod and hscript , even class-wise, so we dont have multiple classes and have to extend to modify and add functionality to the logic like Interp and PolymodInterpEx

This is what I was talking about in my last paragraph of my PR. My goal is to get a PR into Polymod that removes the dependency with the minimal amount of changes possible while guaranteeing no regressions, then a future PR (or ideally even several PRs to split up work into individual, testable changes) can merge classes together to reduce code complexity.

The thing I want to avoid most here is having a rework like this that incidentally breaks something, and we don't notice until much later.

PurSnake:
image
kill dependency

VirtuGuy:
Please Eric. We need this. Polymod isn't supposed to only Friday Night.

NotHyper-474 (approved):
Tested with a mod and works good

NotHyper-474:
Now that this is merged the branch made for the PR can be deleted.

PR #289, opened 2026-01-26
[ENHANCEMENT/FIX] Add extensions and optional fields parsing and fix class syntax parsing for typedefs.

Starexify:
This PR adds correct parsing for optional and extensions in typedefs.

For example this snippet:
```haxe
typedef TestTypedef = {
	var id:Int;
	var data:T;
	var ?version:Float;
}

typedef ExtendedTypedef = {
	> TestTypedef,
	var description:String;
}
```

The ~~hscript~~polymod parser would complain (images below) about the unexpected `>` and `?` symbols before, but now it parses correctly the extensions and optional fields like `version`.

image

image

## Extra
Also fixes the class syntax parsing of typedefs

The example code:
```haxe
typedef TypedefData = {
  public var artist:String;
  private var privates:String;
  //static var renders:Array<String>;
  var strings:Array<String>;
  final finals:Array<String>;
}
```

Before:

image

After:

image

<sub>static context doesnt work in typedefs so I didn't allow it to get parsed<sub>

### DIFF
--- a/polymod/hscript/_internal/Expr.hx
+++ b/polymod/hscript/_internal/Expr.hx
@@ -187,6 +187,7 @@ typedef EnumArgDecl =
 typedef TypeDecl =
 {
   > ModuleType,
+  var extensions:Array<CType>;
   var t:CType;
 }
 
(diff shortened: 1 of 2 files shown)


EliteMasterEric:
Resolves the issue 👍

PR #290, opened 2026-01-26
[FIX] Move operation priority for `is` above other logic operators to match real Haxe behavior.

Starexify:
This PR moves the priority set for `is` above the other logic operators. Which now matches the behavior of real Haxe:

TryHaxe example: https://try.haxe.org/#8BF6C561

Before this would happen:

image

Now it works just like in the haxe example:

image

image


AbnormalPoof (approved):
Yup! That's correct.

Image

PR #291, opened 2026-01-28
Enable processing for all abstracts and typedefs.

EliteMasterEric:
Work in progress to enable processing on every abstract instead of just FlxColor.

Draft because whenever I try to build with this, I get the `(unknown position) Type Not Found: T` error, which implies it's trying to process a generic type somewhere, but since I don't know what type it's trying to process, I can't resolve the issue.

TODO: Try restricting by package, and see if I can narrow down the culprit that way?


NotHyper-474:
I've noticed that explicitly excluding the `haxe.display` package will instead turn the error into a bunch of these, which looks more promising I guess :shrug: 
image

Also, while I'm at it, you might wanna remove the `trace` at line 182 that was added by a previously merged PR.

EliteMasterEric:
> I've noticed that explicitly excluding the `haxe.display` package will instead turn the error into a bunch of these, which looks more promising I guess 🤷

My initial theory is that these are all enum abstracts, and in the compiled code these values get inline replaced, so we'd have to make a separate implementation to handle them.

EliteMasterEric:
The pull request from @lemz1 plus a few additional changes from me have resolved more issues.

You can now access static functions of abstracts, as well as enum values of abstracts.

The only thing you can't really do right now is access instance functions of abstracts, but I'd like to work on that at a future date.

lemz1:
> The only thing you can't really do right now is access instance functions of abstracts, but I'd like to work on that at a future date.

You can still call instance functions like this, although it is a bit hacky

```haxe
var color = FlxColor.RED;
var rgb = FlxColor.get_rgb(color);
```

when trying to call instance functions you do it as if it were a static function, the only thing that you have to do additionally is to parse the instance in the first parameter

so a function like this: 
```haxe
function print():Void {
  trace(this);
}
```

turns into:
```haxe
static function print(this1):Void {
  trace(this1);
}
```

AbnormalPoof (changes requested):
This somewhat works, but

1. You can't import abstract enums
Image
2. Compilation time increases significantly! In Funkin's case it hangs at 100% for around a minute or two

AbnormalPoof (changes requested):
Importing most abstract enums seems to work now! I'm still wondering about the compiler hanging at 100%... what could it possibly be doing there?

NotHyper-474:
Hmm for me the compiler hangs, then its memory usage balloons until it's killed by the system. I can only guess it's something akin to what happened in https://github.com/FunkinCrew/polymod/issues/226.

Then, perhaps the metadata is also getting populated with too much data and thus, generating code that's too "spaghettified" for some compilers to handle?

EliteMasterEric:
I rebased and refactored to make sure more abstracts work, the edge case I found that was broken was one where the abstract implementation class was completely optimized out (due to the class having no static functions that Reflect would want to call). I ended up creating a structure that could be returned (to ensure the game doesn't throw an error that the type couldn't be resolved) that then redirects any property retrieval calls to the appropriate place.

EliteMasterEric:
Just pushed a fix for calling `new FooBar()` when FooBar is an abstract, and also added support for typedefs! You can now just import and use the type without having to think about what the underlying type is called.

polymod/hscript/_internal/PolymodInterpEx.hx
```diff
@@ -1386,7 +1398,12 @@ class PolymodInterpEx extends Interp
     }
 
     // Otherwise, we assume the field is fine to use.
-    if (Std.isOfType(o, PolymodStaticClassReference))
+    if (Std.isOfType(o, PolymodStaticAbstractReference))
+    {
+      // If the target is an abstract, assignment is always invalid.
+      errorEx(EInvalidFinalSet(f));
```
lemz1: this is not true:
if a static variable has a setter, assignment is possible. if we block this here, then the user is forced to `AbstractFoo.set_foo(value)` instead of being able to do `AbstractFoo.foo = value`

[Try Haxe Example](https://try.haxe.org/#Ae3d58C8) (for clarity)

EliteMasterEric:
Just pushed a commit to fix value assignment, still waiting on nothyper to rebase and update the optimization PR so I can add it.

NotHyper-474:
> Just pushed a commit to fix value assignment, still waiting on nothyper to rebase and update the optimization PR so I can add it.

I had already rebased it, then Abnormal reviewed and merged it

EliteMasterEric:
Just rebased this branch and fixed some conflicts, so it'll need more testing.

NotHyper-474:
I'm now getting this error when compiling this branch. Seems like enum abstracts that represent native C++ types don't work very well with the macro. Maybe we should skip extern methods?
image

EliteMasterEric:
@NotHyper-474 I disabled processing `extern abstract` values so please test and let me know if it works.

NotHyper-474:
👍 Builds fine now.

lemz1:
Are there still any things that need to be done for abstracts?

EliteMasterEric:
> Are there still any things that need to be done for abstracts?

I was mostly waiting for a round of approval before merging, since this is a bigger PR.

PR #292, opened 2026-01-28
[FEATURE] String Interpolation

NotHyper-474:
This aims to implement [string interpolation](https://haxe.org/manual/lf-string-interpolation.html) into hscript.

Drafted since it needs some testing (would be appreciated) and fixing nested strings.

### DIFF
--- a/polymod/hscript/_internal/Parser.hx
+++ b/polymod/hscript/_internal/Parser.hx
@@ -46,6 +46,15 @@ enum Token
   TPrepro(s:String);
 }
 
+// NOTE: Refers to string interpolation
+private enum InterpState
+{
+  Literal;
+  Interp;
+  Ident;
+  Expr(depth:Int, inQuote:Bool, ?quoteChar:Int);
+}
+
 class Parser
 {
   // config / variables
@@ -402,6 +411,12 @@ class Parser
         if (e == null) e = mk(EIdent(id));
         return parseExprNext(e);
       case TConst(c):
+        switch (c)
+        {
+          case CString(s, true):
+            return parseExprNext(interpolate(s));
+          default: // Fallback
+        }
         return parseExprNext(mk(EConst(c)));
       case TPOpen:
         tk = token();
@@ -1277,6 +1292,7 @@ class Parser
     readPos = 0;
     allowTypes = true;
     allowMetadata = true;
+    allowJSON = true;
     var decls = [];
     while (true)
     {
@@ -1585,11 +1601,15 @@ class Parser
     return StringTools.fastCodeAt(input, readPos++);
   }
 
-  function readString(until)
+  function readString(until):Const
   {
     var c = 0;
     var b = new StringBuf();
     var esc = false;
+    var interps = false;
+    var maybe = false;
+    var exprDepth = 0;
+    var qt = [false];
     var old = line;
     var s = input;
     #if hscriptPos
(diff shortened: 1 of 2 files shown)


Starexify:
Works almost perfectly, besides ~two issues I found that is different from haxe which are:
1. commas don't work if used between variables in interpolated strings:
So for example I try:
`trace('test $testVar, $testStr');` it outputs as `test testVartestString` instead of the haxe behavior `test testVar, testString`
2. Im not sure if it should but without using {} for functions it shows an Object string like `Object` which doesn't seem to not happen in haxe ?

Which you can see in these images, (1st issue line :10, 2nd issue lines 16/17) :

Polymod:

image

TryHaxe:

image

NotHyper-474:
>  2. Im not sure if it should but without using {} for functions it shows an Object string like `Object` which doesn't seem to not happen in haxe ?

That's just target-specific behavior of trace / Std.string in C++ vs JavaScript, which is outside the scope of this PR.

EliteMasterEric:
Definitely needs thorough testing but this would be awesome to implement.

Of note, the value of `interpreted` is currently just whether the string is using single quotes, maybe it should also check if the string contains a `$` during parsing to potentially save time later on?

NotHyper-474:
> Of note, the value of `interpreted` is currently just whether the string is using single quotes, maybe it should also check if the string contains a `$` during parsing to potentially save time later on?

I've thought of that too, but figured that instead of a single costly loop, there would be two loops instead, one being for `indexOf` and the other for `interpolate`, though perhaps I'm overthinking it.

EDIT: Actually, we can re-purpose `readString` in a way that it checks for the `$` char and returns the `CString`, that should hopefully do the trick.
EDIT 2: Done.

polymod/hscript/_internal/Parser.hx
```diff
@@ -1275,6 +1291,7 @@ class Parser
     readPos = 0;
     allowTypes = true;
     allowMetadata = true;
+    allowJSON = true;
```
NotHyper-474, line 1295: I've staged this change by accident, but perhaps I should keep it, since it would allow parsing of objects with strings for keys (e.g `FlxTween.tween(obj, { "velocity.x": 4 }, ...)`)

AbnormalPoof: that'd be awesome if it was here

NotHyper-474:
I'm happy to say that this is finally ready for testing.

Here's a small assert script I made for this with all the usual cases I could think of:
<details><summary>Assert Script</summary>
<p>

```haxe

import funkin.modding.module.ModuleHandler;

class StrInterpAssert extends Module
{
  function new()
  {
    super("InterpAssert", -1);

    var x = 10;

    function test(s:String)
    {
      return s;
    }

    assertStr("${'{'}", '${'{'}', "{");
    assertStr("${\"{\"}", '${"{"}', "{");
    assertStr("${test('string')}", '${test('string')}', "string");
    assertStr("${test('string ${0}')}", '${test('string ${0}')}', "string 0");
    assertStr("$x and ${test('$x') + ' 10'}", '$x and ${test('$x') + ' 10'}', "10 and 10 10");
    assertStr("$$", '$$', "$");
    assertStr("$$$x ", '$$$x ', "$10 ");
    assertStr("$-=", '$-=', "$-=");
    assertStr("$x ", '$x ', "10 ");
    assertStr("$x", '$x', "10");
    assertStr("$", '$', "$");
    assertStr("$ x", '$ x', "$ x");
    assertStr("$x + 3 is ${x + 3}", '$x + 3 is ${x + 3}', "10 + 3 is 13");
  }

  function assertStr(literal:String, str:String, ass:String)
  {
    if (literal?.length == 0 || str?.length == 0 || ass?.length == 0) throw 'Forgot an argument!';
    trace('ASSERT: ("' + literal + '" -> "' + ass + '"), RESULT: "' + str + ((str == ass) ? '" (SUCCESS)' : '" (FAIL)'));
  }
}
```

</p>
</details>

AbnormalPoof (approved):
Works amazingly! I ran your assert code in an update loop and it worked with little to no performance impact:

Image

Also tried this:
```haxe
FlxTween.tween(this.superClass, {"scale.x": 10}, 5, {ease: FlxEase.expoInOut});
```
Which definitely works!

Image

PR #294, opened 2026-01-29
Implement the `scriptHas` function to check for field existance

### LINKED ISSUE #211: [ENHANCEMENT] Some sort of `hasFunction` function for scripted classes
So you can do something like:
```haxe
if (class.hasFunction('func'))
{
  // logic would go here
}
```
I believe right now you have to directly call the function in a `try/catch` so classes that don't have the function don't crash, which isn't that intuitive.

EliteMasterEric:
Resolves #211.


AbnormalPoof (approved):
Works very nicely! I added a static equivalent so static fields and functions can be checked as well.

PR #297, opened 2026-01-29
[FIX] Fix scripted classes being uninstantiable if the package is empty

lemz1:
example:
```haxe
package;

@:hscriptClass
class ScriptedFlxBasic extends flixel.FlxBasic implements polymod.hscript.HScriptedClass {}
```

before: `hscriptedClasses` metadata contains `.ScriptedFlxBasic`
after: `hscriptedClasses` metadata contains `ScriptedFlxBasic`

this makes `Type.resolveClass` now work correctly


AbnormalPoof (approved):
Very easy fix, works nicely!

PR #300, opened 2026-02-01
[FIX] Pass variables by reference to local function clone 

NotHyper-474:
Instead of copying each variable to the clone, we pass them by reference. This fixes an issue where they are unaffected by local functions.


AbnormalPoof:
300th pr

AbnormalPoof (approved):
Fixes in-game cutscenes being unskippable in Funkin' along with `FlxFlicker` being busted!

PR #303, opened 2026-02-02
[`experimental-abstracts`] Fix compiler stall

NotHyper-474:
This fixes a reported compiler stall that is caused by the resulting C++ code of the `PolymodScriptedClassMacro` meta being too big, by storing the abstract static fields in a [Resource](https://haxe.org/manual/cr-resources.html) instead.

Although, I'm not sure if this PR broke it, but I still can't seem to import some abstracts.
image

### DIFF
--- a/polymod/hscript/_internal/PolymodScriptClassMacro.hx
+++ b/polymod/hscript/_internal/PolymodScriptClassMacro.hx
@@ -18,6 +18,11 @@ using StringTools;
 class PolymodScriptClassMacro
 {
   /**
+   * The name for the Haxe resource that stores the abstract static field paths.
+   */
+  static inline final ABSTRACT_STATICS_RES_NAME:String = 'PolymodScriptClassMacro_AbstractStatics';
+
+  /**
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
I rebased the abstracts branch so you'll have to fix this branch up.

I'm still trying to figure out how best to resolve the issue where some abstracts won't import, but I've narrowed down the cause to the fact that if an `enum abstract` contains no functions (just values), all the values get inlined and then the implementation class gets optimized out entirely. I think I added a function to the `enum abstract` to CharacterRenderType while testing to verify, then forgot I had done that while testing further. I can now reliably reproduce CharacterRenderType not resolving, but figuring out a fix will be annoying considering `AbstractStaticMembers` currently relies on resolving the implementation class to work.

AbnormalPoof (approved):
Improves compile times by a lot! The issue with value-only enum abstracts not being able to be imported was resolved, as I can import `CharacterRenderType` just fine

PR #311, opened 2026-02-04
[FIX] Prevent argument loss in local functions

NotHyper-474:
> [!NOTE]
> #279 introduced a crash which may happen when testing this PR. 

There was already a PR that had addressed this, but it introduced a bug where variables of a class returned to their previous value after a local function ended. I made another PR that fixed that, but at the cost of a regression.

So this is the ultimate fix for this issue, which is as jank as ever. 

Here's a repro code snippet which demonstrates both of the issues addressed.
```haxe
import funkin.modding.module.ModuleHandler;
import flixel.util.FlxTimer;
import flixel.FlxG;


class Test extends Module
{
  function new()
  {
    super("test", -99999);
  }

  var test = 0;

  public override function onUpdate(event:ScriptEvent):Void
  {
    super.onUpdate(event);
    
    if (FlxG.keys.justPressed.ONE)
    {
      example1(() -> {
      	trace("Callback called!");
      }, 474);

      function example2()
      {
        FlxTimer.wait(0, () -> {
          test ++;
          trace('test ${test}');
        });
      }
      example2();

      // Bug: "test" is unaffected by local functions.
      trace('test again: $test');
    }
  }

  // Regression: arguments are lost in native function callbacks
  public function example1(callback:() -> Void, obj:Dynamic):Void
  {
    FlxTimer.wait(1, function():Void {
      callback(); // errors saying it cant find "callback"
      trace(Std.string(obj)); // will also error saying it cant find "obj"
    });
  }
}
```

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -27,6 +27,8 @@ class PolymodInterpEx extends Interp
 
   var _propTrack:Map<String, Bool> = [];
 
+  var _root:Null<PolymodInterpEx>;
+
   function getClassDecl():ClassDecl
   {
     if (_classDeclOverride != null)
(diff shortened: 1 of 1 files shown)


AbnormalPoof:
this is very hotnyper trying not to laugh 👀

Starexify:
> Note
> 
> #279 introduced a crash which may happen when testing this PR.

Yep I messed it up there 😬
- But dont worry I fixed it in  #312 😌

Starexify (approved):

AbnormalPoof (approved):
Works great!

PR #312, opened 2026-02-04
[FIX] Fix crash due to the new param count validation.

Starexify:
This PR fixes a crash introduced in my PR #279 which crashed the game if getters/setters functions are used.

It also adds a custom Error for invalid arguments count.


AbnormalPoof (approved):
Fixes properties, thanks!

PR #314, opened 2026-02-04
[FIX] Fix compiler error when hscriptPos is not defined

lemz1:
We just add some preprocessor checks to ensure variables, that do not exist when hscriptPos is not defined, aren't being assigned to


AbnormalPoof (approved):
Looks correct

PR #317, opened 2026-02-05
Make OpenFL sample use `scriptInit` instead of `init`

NotHyper-474:
I forgot to account for those in my other PR


AbnormalPoof (approved):
lol

Issue #37, opened 2021-09-17
[ENHANCEMENT] Add functions to retrieve files from specific mods, or to ignore mods.


MasterEric:
I was going to submit an issue that was like "hey wouldn't it be cool if you could get the original asset without consideration of any mods installed?" but then I found that `PolymodAssetLibrary.getTextDirectly` and `checkDirectly` literally already exist.

After further investigation, this is actually misleading; "without consideration of any modifications" means without merges or appends. It does NOT allow for the user to forcibly ignore mods.

In most general cases, the game client should only need the instance of the file provided by the last mod in the list that replaces it. However, in special cases, new functions would allow for accessing the original, unmodded file, or even accessing the file from a specific mod. My suggestion is the following set of static functions to be added to the Polymod class (since I don't believe we can modify the method signature of the existing Lime calls):

* `listDirectly(type:PolymodAssetType, modId:String = ""):Array<String>` would list all files of the given asset type. If a value is provided for modId, it would list the files provided by that specific mod. If no value is provided, it will list the asset files only from the fallback libraries, while ignoring mods.
* `getTextDirectly(id:String,PolymodAssetType, modId:String = ""):String` would retrieve the file contents as a string; if a specific modId is specified, the file will be fetched from that, and if no modId is specified, the original unmodified file will be used. 
* `getBytesDirectly(id:String,PolymodAssetType, modId:String = ""):Bytes` is similar to getTextDirectly but it returns bytes.
* `listModsOnFile(id:String)` would list each mod that replaces a file, in mod-load order; thus, the last mod in the list is the mod which is currently providing the file.

MasterEric:
Moved to MasterEric/Polymod#10

Starmapo:
Has any work been started on this yet? If not then I'd be willing to make a PR for it, cause honestly it's the only thing missing from this library that would fully convince me to use it.

EliteMasterEric:
No work has started on this, feel free to look into this if you like.

EliteMasterEric:
The ability to directly retrieve a modded asset from a specific mod, the ability to list assets from a specific mod, and the ability to retrieve/list assets with mods excluded is one of my top most wanted features for Polymod right now.

Starexify:
Id say this should be made alongside a modid system, to use a modid for retrieving files from another mod instead of folder names which can be easily changed by the user when downloading the mod 
(Same should be done with dependencies)

EliteMasterEric:
[...]

PR #330, opened 2026-02-05
[FEATURE] Allow ModMetadata to have an `id` field

Starexify:
This PR adds aims to replace the mod ID system to allow custom IDs for mods instead.
It won't break mods that don't have an `id` field in their metadata, nor their dependencies, because if the id is empty it will default to the directory name like it was in the old behavior. 

> [!IMPORTANT]
> If this PR is accepted, games like Friday Night Funkin' that use the polymod framework should change from `meta.id` to `meta.dirName` when loading mods.

### Old Problem
With the old system mods could ship zips with different names other than the dependency name other mods are using, for example if the dependent mod's metadata had:
```json
{
  "dependencies": {
    "Mod": "0.0.0"
  }
}
```

Polymod would check for the folder name `Mod`, but a user, the modder or some platform could ship the mod folder name differently like `Mod-main` or `Mod-b010349` instead inside the archive, and some people extract archives without the archive's name and then need to create one for the mod and then extract the contents there.

### New Problem
A mod id can be used between multiple mods, so a conflict check should be added in the environment that loads mods. But the chance of that happening is very low.

### Extra
This PR also adds a new function in `FileSystem` named `getMetadataById(modId:String)`, which allows the recieving of metadatas based on the ID of a mod. Also renames the `getMetadata(modId:String)` to `getMetadataByDir(dir:String)`. **(only for SysFileSystem at the moment!)**

image

### DIFF
--- a/polymod/Polymod.hx
+++ b/polymod/Polymod.hx
@@ -253,15 +253,15 @@ class Polymod
     {
       if (dirs[i] != null)
       {
-        var modId = dirs[i];
-        var meta:ModMetadata = fileSystem.getMetadata(modId);
+        var dir = dirs[i];
+        var meta:ModMetadata = fileSystem.getMetadataByDir(dir);
 
         if (meta != null)
         {
           if (!VersionUtil.match(meta.apiVersion, params.apiVersionRule))
           {
             error(MOD_API_VERSION_MISMATCH,
-              'Mod "${modId}" was built for incompatible API version ${meta.apiVersion.toString()}, expected "${params.apiVersionRule.toString()}"', INIT);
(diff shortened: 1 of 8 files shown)


TechnikTil (approved):
yes, i need this.

Starexify:
Will mention in comments that #37 should benefit from a better modId system like this if the assets will be inside `modIds` instead of simply like: `assets/dir/script.hxc` to be `assets/modId/script.hxc`.
But not very sure about this approach, would need more thoughts on this

requazar (approved):

Trofem (approved):
I feel kinda id-less

AbnormalPoof:
It's time.

AbnormalPoof (approved):
Dude it's so fucking peak

Image

PR #332, opened 2026-02-07
[FEATURE] Add support for `import.hxc` files

KoloInDaCrib:
Closes https://github.com/FunkinCrew/polymod/issues/329

This PR adds support for `import.hxc` files that add imports and usings to the modules (not to be confused with FNF Modules) that have the same or a child package. This does not add imports and usings based on the folder the scripts are in, considering scripts can be placed anywhere in a mod and have their modules loaded.

image
image

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -267,6 +267,10 @@ class PolymodInterpEx extends Interp
     // Clear the script class descriptors.
     _scriptClassDescriptors.clear();
 
+    // Also clear the imports from the import.hx files.
+    _scriptClassImports.clear();
+    _scriptClassUsings.clear();
+
     // Also destroy local variable scope.
     this.resetVariables();
   }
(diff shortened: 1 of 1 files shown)


Starexify:
So checklist:
1. What if the `import.hxc` is loaded AFTER the hxc scripts were initialized, would it work ?
2. Does it use the `package` from the `import.hxc` to determine what scripts have the default imports ? (in haxe you cant do that, i'd do it using the folder location but I guess this works with the way polymod handles packages aswell)
3. Does having a file named `class-import.hxc` see it as an `import.hxc` file?
4. Do aliases work ?

KoloInDaCrib:
> So checklist:
> 
> 1. What if the `import.hxc` is loaded AFTER the hxc scripts were initialized, would it work ?
> 2. Does it use the `package` from the `import.hxc` to determine the package it should go in ? (in haxe you cant do that, i'd do it using the folder location but I guess this works with the way polymod handles packages aswell)
> 3. Does having a file named `class-import.hxc` see it as an `import.hxc` file?
> 4. Do aliases work ?

1. Imports from both scripted classes and now import.hxc are validated after all scripts are loading, hence the code is in `validateImports`
2. Yes, it uses packages. I didn't want to use folder locations because you can put scripts anywhere, as opposed to being limited to putting .hx files in the source folder of haxe projects.
3. No.
4. Yes.

Starexify (approved):

AbnormalPoof (approved):
works great

PR #335, opened 2026-02-08
[FIX] Properly account for property access from outside the class

### LINKED ISSUE #308: Getters and Setters function inconsistently compared to real Haxe
I was working on an Alphabet class for a mod in FNF, and with the recent ability to extend scripted classes I thought it'd be a good time to split typed text to its own class. 

There is a variable "text" in the base class which has an associated `set_` function to reset a few things, but the typed text would require a few extra things reset as well. I found out though that the overriden `set_text` function doesn't seem to called at all in the typed text's class, so I looked a little bit into it.

With the code below, running it through hscript will only show 4 traces, but running it in the Try Haxe site will show all 8 expected traces.

Image

Image

https://try.haxe.org/#75f780f4

```haxe
class TestPlayground
{
    // call this from somewhere else
    static function main():Void
    {
        trace("--TestThing--");
		var thing:TestThing = new TestThing();
		var getvalue:Int = thing.test;
		thing.test = 10;

        trace("--TestThing2--");
		var thing2:TestThing2 = new TestThing2();
		var getvalue2:Int = thing2.test;
		thing2.test = 100;
    }
}

class TestThing
{
	@:isVar
	public var test(get, set):Int = 0;

	public function new():Void {}

	function get_test():Int
    {
		trace("get_test");
		return test;
	}

	function set_test(value:Int):Int
    {
		trace("set_test");
		test = value;
		return value;
	}
}

class TestThing2 extends TestThing
{
	override function get_test():Int
    {
		trace("- get_test #2");
		return super.get_test();
	}

	override function set_test(value:Int):Int
    {
		trace("- set_test #2");
		return super.set_test(value);
	}

	public function new():Void
    {
		super();
	}
}
```

NotHyper-474:
Closes #308 

Fixes accessing a variable of a class, from another class, not accounting for properties correctly.
There's repro code in the linked issue.

### DIFF
--- a/polymod/hscript/_internal/PolymodAbstractScriptClass.hx
+++ b/polymod/hscript/_internal/PolymodAbstractScriptClass.hx
@@ -29,7 +29,7 @@ abstract PolymodAbstractScriptClass(PolymodScriptClass) from PolymodScriptClass
         }
         else if (this.findVar(name) != null)
         {
-          var v = this.findVar(name);
+          var v = this.findVar(name, true);
 
           @:privateAccess
           switch (v.get)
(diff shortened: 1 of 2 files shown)


NotHyper-474:
Before I forget it, merging this will cause a crash on Freeplay when the BF cartoon cycle starts since the script tries to modify a scripted final (which should not be allowed), and this PR accidentally fixes that never giving an error. The BF DJ code has to be changed to account for this.

AbnormalPoof (approved):
Tested with the repro code in the issue, works great!

PR #343, opened 2026-02-10
Fix persistence of named local functions

NotHyper-474:
Fixes https://github.com/FunkinCrew/polymod/issues/322

This came from the original Interp code, which never really had a class design in mind.


NotHyper-474:
Superseded by https://github.com/FunkinCrew/polymod/pull/342 (I had made the same fix there and forgot to remove it :EricWhoopsie:)

PR #344, opened 2026-02-11
[REFACTOR] Fold `PolymodExprEx` into `Expr`

NotHyper-474:
Removes `PolymodExprEx` and `PolymodPrinterEx`, then merges the former's code into `Expr` while also changing the code to account for the changes. This unifies the error handling for `ErrorEx` and `Error`, resulting in way cleaner and less repetitive code.

From my little bit of testing, the only functional change this has is that now some error messages don't display `hscriptClass(path/to/file.hxc):676: Error Message` anymore, which I suppose is fine since `PolymodScriptClass.reportError`, albeit not including the file of origin, still includes the class name, function name, and line number.


EliteMasterEric (approved):
Great refactor, big improvement to the codebase, Funkin' builds and doesn't break scripts.

PR #345, opened 2026-02-12
Fix another compiler error when `hscriptPos` is not defined

NotHyper-474:
> [!NOTE]
> This PR will not fully solve the error because the fix from https://github.com/FunkinCrew/polymod/pull/314 went to `experimental-abstracts` instead, and I don't wanna apply the fix that was already done there.

It is important to know that `Expr` changes depending of if `hscriptPos` is enabled; this PR fixes a function not accounting for that.

P.S: This file needs another reformat :obese_cat:


Starexify:
Oh forgot to format that to Allman style my bad 😬
also didn't think about `hscriptPos` since I dont use it and not sure what it does uf

PR #347, opened 2026-02-17
Fix a couple of issues in the changelog entries

NotHyper-474:
> [!NOTE]
> PR based on `experimental-abstracts` since a new changelog commit went there as well. 

- Fixes a contributor not being properly credited.
- Moves the denotion of `PolymodConfig.caseInsensitiveZipLoading` to the sub-entry of the PR that was actually responsible for adding it.


Hundrec:
Actually I'll just merge it myself I love changelogs

PR #349, opened 2026-02-17
[ENHANCEMENT] Add `scriptInit` and `listScriptClasses` to every script class

KoloInDaCrib:
Since scripted classes can extend each other now, it is a no-brainer to add functions from normal script class bases into scripted classes.
I initially wanted to add onto the staticFields array when initializing a class, however I couldn't figure out how to convert the functions into `hscript.Expr` blocks without using a different interpreter for them, as I would need some of Polymod's classes which are normally blacklisted.

image

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -1723,6 +1723,50 @@ class PolymodInterpEx extends Interp
    */
   public function callScriptClassStaticFunction(clsName:String, fnName:String, args:Array<Dynamic> = null):Dynamic
   {
+    // For functions listScriptClasses and scriptInit, we want to return the needed values without much checking.
+    if (fnName == "listScriptClasses")
+    {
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
Oh I see, you're looking to dynamically instantiate scripts from ANOTHER script! I was going to suggest using the `constructor` but this is a different thing.

As for the actual implementation, this could probably be handled using a single `Map<String, Function>`, I could see a bug happening where a function gets added in one place and not the other, causing issues.

AbnormalPoof (commented):
I'm not sure if this is correct, but the code for `scriptInit` could imply that you can instantiate ANY class instead of only ones that extend said class.

KoloInDaCrib:
> I'm not sure if this is correct, but the code for `scriptInit` could imply that you can instantiate ANY class instead of only ones that extend said class.

This is true, as the code for this PR was taken directly from how `scriptInit` is handled by the script class bases. The only difference between the two is that the script class bases force a return type.
image

AbnormalPoof (approved):
Works great!

PR #364, opened 2026-03-03
[FEATURE] Add the ability to merge scripts using `_merge` 

KoloInDaCrib:
Basically just a rework of https://github.com/FunkinCrew/polymod/pull/236 but with features that were included with the recent polymod updates and without a potential for a horrible merge conflict :obese_cat:

This time I've summed up the system in a [coolbool](https://github.com/KoloInDaCrib/code-cookbook/blob/script-merging/assets/content/cookbook/Expert/06.ScriptMerging.md) article that I'm planning to push to the repo when this gets merged.

### DIFF
--- a/polymod/format/ParseRules.hx
+++ b/polymod/format/ParseRules.hx
@@ -31,6 +31,7 @@ class ParseRules
       case JSON: new JSONParseFormat();
       case LINES: new LinesParseFormat(EndLineType.LF);
       case PLAINTEXT: new PlainTextParseFormat();
+      case SCRIPT: new ScriptParseFormat();
       default: new PlainTextParseFormat();
     }
     formats.set(extension, format);
@@ -59,6 +60,13 @@ class ParseRules
     rules.addFormat('xml', new XMLParseFormat());
     rules.addFormat('json', new JSONParseFormat());
     rules.addFormat('txt', new PlainTextParseFormat());
+
+    rules.addFormat("hxc", new ScriptParseFormat());
+    for (ext in PolymodConfig.scriptClassExt)
+    {
+      rules.addFormat(ext, new ScriptParseFormat());
+    }
+
     return rules;
   }
 }
@@ -697,6 +705,7 @@ enum TextFileFormat
   TSV;
   XML;
   JSON;
+  SCRIPT;
 }
 
 enum EndLineType
--- a/polymod/format/ScriptParseFormat.hx
+++ b/polymod/format/ScriptParseFormat.hx
@@ -1,0 +1,410 @@
+package polymod.format;
+
+import polymod.format.ParseRules;
+import polymod.hscript._internal.Expr;
+import polymod.hscript._internal.Parser;
+import polymod.hscript._internal.Printer;
+
+/**
+ * A parse format for scripts. While Imports and Usings automatically get added and Packages get ignored,
+ * the metadata is used to determine how Modules (Classes, Typedefs, etc.) are handled.
+ * - `@:merge_add` - Directly add the Module, if one with the same name doesn't exist. Also works on Class Fields.
+ * - `@:merge_override` - Override the Module, if one with the same name exists. Otherwise, add it normally. Also works on Class Fields
+ * - `@:merge_combine` - Combine the fields of the two Modules, if the base one exists. Otherwise, add it normally. Only works on Classes.
+ * - `@:merge_insert(index)` - Insert the expression into the base class at a specified index. Only works on Functions.
+ * If a Module has more than one metadata, only the one with the highest priority is used, with the order being Combine > Insert > Add > Override.
+ */
+class ScriptParseFormat implements BaseParseFormat
+{
+  public static final ADD_META:String = ":merge_add";
+  public static final COMBINE_META:String = ":merge_combine";
+  public static final OVERRIDE_META:String = ":merge_override";
+  public static final INSERT_META:String = ":merge_insert";
+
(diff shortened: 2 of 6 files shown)


TechnikTil:
we are so back

TechnikTil:
you know you couldve just force pushed the rewrite right

KoloInDaCrib:
> you know you couldve just force pushed the rewrite right

:Perhaps:

EliteMasterEric:
I'm curious how this is different/better than the existing PolyMixin library, which seems to provide more precise injection while also not even requiring changes to Polymod: https://starexify.github.io/BadDocs/src/polymixin.html#injection-params

Starexify:
> I'm curious how this is different/better than the existing PolyMixin library, which seems to provide more precise injection while also not even requiring changes to Polymod: https://starexify.github.io/BadDocs/src/polymixin.html#injection-params

Well while working on PolyMixin I encountered some challenges, one is that it only works because of V-Slice (the base game) since it caches the AST of scripts into the `ModuleHandler` and other Registry classes like `NoteKindsRegistry` and such. Now this **COULD** potentially be done better if access to all parsed AST scripted classes/enums etc is given by the game or polymod itself (not yet sure if polymod caches all of them somewhere in form of AST or not, I didn't dive that deep).

Now on this PR, it does provide some way of doing mixins similarly to my library in form of merging which is what PolyMixin lacks because of what I stated above about limitations with cache. And for me it looks like it has potential to be expanded in the future with more features like PolyMixin if needed. It might also be better for modders to use the builtin polymod mixin capabilities provided by this PR rather than downloading a library like PolyMixin which indeed provides more functionality (at the moment) for mixins but isn't perfected and again has limitations.

So personally I would either expand on this PR after being merged with more functionality similar to PolyMixins, or if not, then polymod (or the game) could provide some kind of `postParse` event to allow any mod to inject into the parsed AST directly before all scripts are loaded, which would make it easier and better to mixin via mods. Because PolyMixin is limited to and by V-Slice as of now.


On a personal note, I would change the meta naming, so instead of `@:merge_insert(index)` to use `@Inject(ordinal = i)` but that is because I prefer the SpongePowered/Java style. But this can be discussed later best.


If there are any other questions about this, I'll be glad to answer.

VirtuGuy:
This would be so awesome.

AbnormalPoof:
> I'm curious how this is different/better than the existing PolyMixin library, which seems to provide more precise injection while also not even requiring changes to Polymod: [starexify.github.io/BadDocs/src/polymixin.html#injection-params](https://starexify.github.io/BadDocs/src/polymixin.html#injection-params)

Like Starexify said, it *only* works within the confines of FNF due to the way that FNF structures it's modding system. Making changes to Polymod directly would allow it to work within other projects that use Polymod, along with it likely being a lot more stable since it wouldn't need to do hacky workarounds to be able to modify scripts (which would probably break after a major update anyway).

AbnormalPoof (approved):
This is such an intuitive and amazing system, I believe modders are going to be delighted to see this since it removes the need to completely override scripts, which could conflict with mods.

This is the script I used during testing:
```haxe
import funkin.play.stage.Stage;

@:merge_combine
class MainStageErectStage extends Stage
{
  @:merge_add
  var FUCKMODE:Bool = false;

  @:merge_add
  var modeofFUCK:FuckMode = FuckMode.FUCK;

  @:merge_insert(1)
  override function addCharacter(character:BaseCharacter, charType:CharacterType):Void
  {
    character.useRenderTexture = true;
    character.blend = 11;

    #if desktop
    trace('woo');
    #end
  }

  @:merge_insert(2)
  function onBeatHit(event:SongTimeScriptEvent):Void
  {
    if (this.getBoyfriend() != null)
    {
      this.getBoyfriend().visible = !this.getBoyfriend().visible;
    }

    this.getDad().skew.x += 0.5;
  }

  @:merge_insert
  override public function onCountdownStart(event:CountdownScriptEvent):Void
  {
    super.onCountdownStart(event);

    FUCKMODE = false;
  }

  @:merge_add
  override public function onCountdownEnd(event:CountdownScriptEvent):Void
  {
    super.onCountdownEnd(event);

    FUCKMODE = true;
  }

  @:merge_add
  override public function onUpdate(event:UpdateScriptEvent):Void
  {
    super.onUpdate(event);

    if (FUCKMODE)
    {
      this.getGirlfriend().skew.x += 1;
      this.getGirlfriend().skew.y += 1;
      this.getGirlfriend().angle += 1;
    }

    trace(modeofFUCK);
  }
}

@:merge_add
enum FuckMode
{
  FUCK;
}
```

Which parses correctly!

Image

I also went ahead and made a test script for alt instrumentals, since I recall modders having a lot of trouble with this:
```haxe
import funkin.play.song.Song;
import funkin.save.Save;

@:merge_combine
class BopeeboSong extends Song
{
  @:merge_override
  public override function listAltInstrumentalIds(difficultyId:String, variationId:String):Array<String>
  {
    return ['bf'];
  }
}
```

... Which also works!

Image

It might seem weird at first having to explicitly define metadata even when it doesn't seem necessary, but considering this is inspired by the JSON Patch format, I can give it a pass!

My verdict is that this is **good to merge**, any other small bugs we come across with this can be fixed very easily. This isn't very risky to merge since it only *adds* things instead of removing or modifying other parts of Polymod... worst case scenario is that something just straight up doesn't work.

Mergin' it!!

trayfellow:
images(12)

PR #366, opened 2026-03-03
[BUGFIX] Move `PolymodScriptClassMacro` metadata into a resource

TechnikTil:
## What does this PR do?
This PR moves the registry generated by `PolymodScriptClassMacro` into a `haxe.Resource`. 

This was done due to the C++ file becoming really long in some repositories, making compilers hang and sometimes causing Segmentation faults.

This is what `PolymodScriptClassMacro.cpp` looks like in Friday Night Funkin' currently.
image
With this PR, those lines are gone.

### DIFF
--- a/polymod/hscript/_internal/PolymodScriptClassMacro.hx
+++ b/polymod/hscript/_internal/PolymodScriptClassMacro.hx
@@ -7,7 +7,6 @@ import haxe.macro.Type;
 import haxe.macro.Type.ClassType;
 import polymod.util.MacroUtil;
 #end
-import haxe.rtti.Meta;
 
 using StringTools;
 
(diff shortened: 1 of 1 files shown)


polymod/hscript/_internal/PolymodScriptClassMacro.hx
```diff
@@ -180,17 +184,26 @@ class PolymodScriptClassMacro
...
 
+    var metaData = {
+      hscriptedClasses: hscriptedClassEntries,
+      abstractImpls: abstractImplEntries,
+      typedefs: typedefEntries
+    };
+
+    var metaDataHXSF = haxe.Serializer.run(metaData);
+    Context.addResource(METADATA_RESOURCE_NAME, haxe.io.Bytes.ofString(metaDataHXSF));
+
```
TechnikTil, line 206: Just so you know, I moved this above the `Sys.time` call due to `haxe.Serializer` sometimes being slow, which is something that should be reported with this macro instead of just random compiler jank.

moondroidcoder (approved):
It works !

Hundrec (approved):
It does!

PR #368, opened 2026-03-06
[ENHANCEMENT] Fill the `preprocesorValues` map with defines

KoloInDaCrib:
This PR just adds the values from `haxe.macro.Context.getDefines();`, as the logic was coded in beforehand

Tested with both
```haxe
#if desktop [...] #end
```
and
```haxe
#if (polymod >= "1.0.0") [...] #end
```
and they seem to work fine.


EliteMasterEric:
This is surprisingly simple code for something that achieves something so useful, and in a way that identically matches how the same thing would be done in source! Will review and approve if it works good.

PR #370, opened 2026-03-07
[FIX] Move preprocessor defines in DefineUtil to fix compiler errors.

Starexify:
This PR moves the Define map getter for preprocessors in `DefineUtil` since compilers might error because of it.

Like here:

image


TechnikTil (approved):
yay! i can compile to ios again

PR #371, opened 2026-03-08
[ENHANCEMENT] Add mods to preprocessors

KoloInDaCrib:
An addition to the preprocessor support PR that was made now since I was waiting for Starexify's PR that adds the mod id support to get merged.

image

This is useful for mods that use dependencies and especially import classes from those dependencies.


Starexify (approved):
Love how this could be used alongside dependencies 👍

PR #372, opened 2026-03-11
[FIX] Allow parsing of dotted defines in preprocessor checks

NotHyper-474:
Allows for the parsing of defines like `target.name`. This can be useful to check for the existance of mods like Keoiki's "Better Alphabet", which has dots in its `id`.


AbnormalPoof (approved):
Works

PR #375, opened 2026-03-17
[FEATURE] Add support for Function Bindings

KoloInDaCrib:
Partially closes https://github.com/FunkinCrew/polymod/issues/327 since the issue used wildcards, whereas this implementation doesn't, as any usage of wildcards isn't yet programmed in Polymod.

image


bopel-maki-macohi:
Oh, that's how those work

PR #376, opened 2026-03-18
[FEATURE] `hscript` to Polymod import redirection

NotHyper-474:
## Wat dis do?

Adds a define, `POLYMOD_REDIRECT_HSCRIPT` that when enabled will redirect imports to `hscript` to the internal implementation of Polymod. This can be useful, for instance, to allow the HaxeFlixel Debugger console to work without the need of having hscript installed.

> [!NOTE]
> Building may break in the future the more the classes are merged together.

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -2,7 +2,8 @@
 <project>
   <!-- These values are added to the project XML when you include this library. -->
   <haxelib name="thx.semver" />
-  <!-- <haxelib name="hscript" /> This library is now OPTIONAL. -->
+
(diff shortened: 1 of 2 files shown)


EliteMasterEric (changes requested):
This is pretty neat, but if you install Polymod and don't have `POLYMOD_REDIRECT_HSCRIPT` enabled you get errors that make projects fail to build.

Image

NotHyper-474:
Finally got around to fix that as well, the issue was that you had hscript installed and Polymod was shadowing its files, so I had to scrap the original idea and replace it with a macro instead.

The way it works is quite clever, we just ask Haxe to give us the references that lead to a missing hscript type and replace those with a `typedef` to a Polymod type. I have to thank `Context.onTypeNotFound` for even allowing this to be possible.

PR #377, opened 2026-03-22
[CHORE] Merge the `PolymodParserEx` into `Parser`.

Starexify:
Just merges the `PolymodParserEx` class into `Parser` to remove the redundant extra class, nothing more.


EliteMasterEric:
code looks good and compilation doesn't break 👍

PR #378, opened 2026-03-22
[CHORE] Merge the `PolymodEnumDeclEx` into `EnumDecl`.

Starexify:
Just merges the `PolymodEnumDeclEx` class into `EnumDecl` to remove the redundant extra typedef, nothing more.


EliteMasterEric (approved):
Code looks good and doesn't break compilation 👍

PR #380, opened 2026-03-22
[ENHANCEMENT] Allow single-expression (braceless) functions to be parsed.

### LINKED ISSUE #319: [FEATURE] Support for single-expression (braceless) functions
In haxe, we can do this to have only a single expression from a function, without needing to encapsulate it within braces `{}` :
```haxe
public static function get_y() return 1; // <-- errors here in polymod

static function main() {
  trace(get_y());
}
```

And in a haxe environment it would return:
```
1
```

But in polymod, we get this:
```
ERROR  Error while parsing script assets/scripts/WorstCParser.hxc#5: EUnexpected
Unexpected error: Unexpected token ";", is there invalid syntax on this line?
```

Starexify:
## Linked Issues
- Closes #319

## Description

This PR allows single-expression (braceless) functions to be parsed and not throw errors.
All with a one-liner.

## Screenshot

image

*Note: Local functions such as `lambda()` in this case, worked before this change, module scoped functions didn't*

### DIFF
--- a/polymod/hscript/_internal/Parser.hx
+++ b/polymod/hscript/_internal/Parser.hx
@@ -1657,6 +1657,7 @@ class Parser
         case "function":
           var name = getIdent();
           var inf = parseFunctionDecl();
+          maybe(TSemicolon);
           return {
             name: name,
             meta: meta,


Starexify:
Also here is a tweak that won't work in haxe *(the 2nd function)*:

image

PR #391, opened 2026-04-15
[FIX] Apply expression typing for class fields

NotHyper-474:
Allows Map class fields to properly initialize as maps when using an empty array declaration (e.g `public var map:Map<... , ...> = [];`) . Previously this only happened with local variables.

This happened to break a mod's script, but it was due to user error since it was initializing an `Array<String>` with a Map body.


requazar (approved):
🧩

EliteMasterEric (approved):
Code looks good.

PR #392, opened 2026-04-17
[BUGFIX] Add missing ModMetadata info for `getMetadataById`

ComedyLost:
This pr adds some missing values that weren't being added when getting metadata info using `getMetadataById` such as a mod's directory name, mod path, icon, and icon path

Draft for now until my game can ACTUALLY COMPILE RAHHHHHHHHH


ComedyLost:
was finally able to test it and it works

PR #394, opened 2026-04-19
[FIX] Properly check for ignored files

NotHyper-474:
Allows ignored file patterns like `.git` to properly get ignored.

### DIFF
--- a/polymod/backends/PolymodAssetLibrary.hx
+++ b/polymod/backends/PolymodAssetLibrary.hx
@@ -84,7 +84,7 @@ class PolymodAssetLibrary
   // Cache for directory listings to avoid repeated file system scans
   private var _dirCache:Map<String, Array<String>> = new Map();
   // Fast lookup for ignored files using Map instead of array searches
-  private var _ignoredFilesSet:Map<String, Bool> = null;
+  private var _ignoredFilesCache:Map<String, Bool> = new Map();
   // Cache for file existence checks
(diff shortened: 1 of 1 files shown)


MightyTheArmiddilo:
YOU THE GOAT HYPER!!!!

Raltyro:
instant merge, much improvement, much performance, much speed, much fixes

VirtuGuy:
How much does this benefit performance?

NotHyper-474:
> How much does this benefit performance?

I don't know. Could actually be slightly worse than currently, considering the cache is lazily initialized (we have to iterate through the ignored file names before we cache the result.)

NotHyper-474:
Decided to profile it and the performance loss would be probably unnoticeable.

| Before | After |
|--------|--------|
| image | with-fix | 

We can see it has a slow start, but then gradually it quickly is on par with the times we get without the PR. And we're talking about microseconds, which is too small of a measurement to be perceptible. No performance gains though :obese_cat:

Raltyro:
But obviously it had a degrade on performance with the original commented codes

EliteMasterEric:
The original code had a MASSIVE performance impact (even with caching) because it was using regular expressions to try to implement `glob` rather than simply checking the prefix.

I think having it merely be a prefix is a good compromise for now, but it does prevent stuff like filtering all files with the `.md` extension.

PR #395, opened 2026-04-22
[FEATURE] Add `cast`

gamerbross:
I've been doing some trickery to have autocompleting on code for scripts and needed to cast stuff so no errors happen.

They don't really have a functionality, just parses them and interpret the expression cus hscript doesn't have types

Supports `cast object`, `cast(object)` and `cast(object, CoolClass)`

I'm not sure about what I did in Printer.hx and Tools.hx so
any suggestions are welcome!! :  )

### DIFF
--- a/polymod/hscript/_internal/Expr.hx
+++ b/polymod/hscript/_internal/Expr.hx
@@ -61,6 +61,7 @@ enum Expr
   EFor(v:String, it:Expr, e:Expr);
   EBreak;
   EContinue;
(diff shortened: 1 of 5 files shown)


EliteMasterEric:
Code looks good, just need to make sure it doesn't break existing scripts.

EliteMasterEric:
This is a good PR even without the autocomplete stuff, because before if you copy-pasted a chunk of code that happened to have a `cast()` in it it would throw an error and now it doesn't

PR #398, opened 2026-04-23
[FIX] Allow imports with module names to work

NotHyper-474:
This PR makes it so doing imports like `foo.bar.Waldo.Fred` work instead of having to write it as `foo.bar.Fred`. The fix is such a no-brainer which surprises me someone hasn't tried to do it yet.

Note that this technically makes it so putting anything as the module will work, but I don't think there's a way to validate that (we have enough macros already) and I don't think it's really worth it since this PR just makes it more convenient to use hscript with auto-completion.

The scary diff is from me trying to avoid boilerplate code.

### DIFF
--- a/polymod/hscript/_internal/PolymodInterpEx.hx
+++ b/polymod/hscript/_internal/PolymodInterpEx.hx
@@ -2154,6 +2154,59 @@ class PolymodInterpEx extends Interp
     }
   }
 
+  /**
+   * Tries to resolve the type of an imported class, which will end up in `cls`, `enm` or `abs`.
+   * @param importedClass The import to resolve.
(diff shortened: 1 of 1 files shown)


EliteMasterEric (changes requested):
I happened to have `import haxe.Unserializer;` in a script where I was testing blacklist stuff.

Before this PR, it silently fails to import, then throws an error when you try to access the blacklisted module when you call a function that uses the import.

After this PR, it throws an error when trying to import (one that doesn't mention the fact the module is blacklisted).

Image

Either revert to the previous behavior (script error thrown when calling the blacklisted class) or add an error when attempting to import the blacklisted class.

NotHyper-474:
Fixed by reverting to the previous behavior.

AbnormalPoof (approved):
Works

PR #399, opened 2026-04-24
[REFACTOR] Clean up `HScriptedClassMacro`

NotHyper-474:
This PR aims to improve the `HScriptedClassMacro` class readability by improving the usage of features of the Haxe language throughout the code.

* Removes unused local variables and code.
* Applies the use of `Context.follow` to get rid of `TLazy` (constructors don't like this though, so not for them...)
* Adds the needed imports for the types it uses and replaces the use of the full module name where those are referenced.
* Simplifies the checks for metadata.
* Moves the check for `display` mode to before the field overriding and after the addition of the basic functions (`scriptInit`, `scriptGet`, etc.) to prevent issues with errors about those not being found on the language server.
  - Screenshot included Captura_de_tela_20260412_221302

* Improves the usage of pattern matching in switch cases to reduce verbosity and/or boilerplate.

### DIFF
--- a/polymod/hscript/_internal/HScriptedClassMacro.hx
+++ b/polymod/hscript/_internal/HScriptedClassMacro.hx
@@ -1,7 +1,11 @@
 package polymod.hscript._internal;
 
+#if macro
 import haxe.macro.Context;
 import haxe.macro.Expr;
+import haxe.macro.Expr.ComplexType;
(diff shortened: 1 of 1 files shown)


VirtuGuy:
pico surprised

requazar:
peak

PR #400, opened 2026-04-28
[FEATURE] Include Case-Insensitive FS lookup for almost all operating systems

TechnikTil:
## What does this PR do?
This PR changes all preprocessor checks for `getPathLike` to be `!windows` instead of `linux`.
Linux is not the only case-sensitive operating system, so this PR effectively adds this feature to almost all operating systems that also have case-sensitivity.

### DIFF
--- a/polymod/fs/SysFileSystem.hx
+++ b/polymod/fs/SysFileSystem.hx
@@ -8,7 +8,7 @@ import polymod.util.Util;
 import polymod.util.VersionUtil;
 import thx.semver.VersionRule;
 
-#if linux
+#if (!windows)
 using StringTools;
 #end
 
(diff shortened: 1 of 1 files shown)


MightyTheArmiddilo:
This might be huge 👀
###### I say as a Windows user

moxie-coder (approved):
I approve as an Linux user

###### and a Windows user

NotHyper-474:
400th PR :eyes:

###### also I'm pretty sure you don't need the parenthesis for the check

NotHyper-474:
Just realized this didn't account for the override in `SysZipFileSystem`, now ZIP mods are probably broken on macOS and iOS 

image

PR #401, opened 2026-05-01
[BUGFIX] Fix Scripted Class Extending not working with multiple scripted layers

TechnikTil:
## What does this PR do?
There is currently a bug with `HScriptedClassMacro` where after multiple layers of scripted class extending, it would not detect any layers except the first two. This PR fixes that by making the super lookup recursive.

## More Info about the Bug
By applying [this patch](https://github.com/user-attachments/files/27283626/scripted-class-layer-bug.patch), the `openfl_hscript_class` example will include two new stages in `mod8`. "Even Less Basic Stage" should not work unless this PR is the base. This is due to it extending "Less Basic Stage", which extends "Basic Stage".


TechnikTil:
Updated the code to use less lines

PR #402, opened 2026-05-03
Implement `hscriptPos` to be fully built in

realvirtu:
This PR removes the need for `hscriptPos` in order for line numbers to be used in script errors. `hscriptPos` doesn't need to be something you have to manually enable. In fact, you probably never want it disabled, because why wouldn't you want line numbers??

This should also put an end to PRs having to be made due to any compiling problems that occur because of `hscriptPos` being disabled.

### DIFF
--- a/polymod/format/ScriptParseFormat.hx
+++ b/polymod/format/ScriptParseFormat.hx
@@ -51,7 +51,7 @@ class ScriptParseFormat implements BaseParseFormat
     }
     catch (e:Error)
     {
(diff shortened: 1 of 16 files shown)


Starexify:
hscriptPos should be kept, as it makes polymod not to be bloated and slow from traces, you dont always need the trace pos in some projects.

realvirtu:
image

EliteMasterEric:
Agreed tbh. It shouldn't error if it's disabled but also it should be opt-in since there's technically a performance impact.

PR #411, opened 2026-05-25
Fix HashLink crash

NotHyper-474:
Fixes a crash about an invalid cast on HashLink.

The compiler guesses the type of `_super` as of `PolymodScriptClass`, but it can be of the instance of a `HScriptedClass` extending scriptable class. We force it to type as `Dynamic` instead.


NotHyper-474:
thanks as always ashnormal

PR #415, opened 2026-06-10
Re-enable checks for final access.

EliteMasterEric:
- Improve error messages for invalid access to `final` variables in scripts.
- Improve handling of properties with `null` or `never` setters (Haxe never allows assignment to `never`, and only allows local assignment to `null`).
- Re-enable the code to prevent assignment to `final` constants from Haxe code. This prevents issues like being able to re-assign values despite them being defined as immutable constants.

### DIFF
--- a/checkstyle.json
+++ b/checkstyle.json
@@ -172,7 +172,9 @@
         "ignoreOverride": true,
         "tokens": ["ABSTRACT_DEF", "CLASS_DEF", "ENUM_DEF", "INTERFACE_DEF"],
         "modifier": "PUBLIC",
-        "excludeNames": ["new", "toString"]
+        "excludeNames": ["new", "toString"],
+
+        "severity": "IGNORE"
       },
       "type": "FieldDocComment"
     },
@@ -494,7 +496,9 @@
       "props": {
         "enforceReturnTypeForAnonymous": false,
         "allowEmptyReturn": true,
-        "enforceReturnType": true
+        "enforceReturnType": true,
+
+        "severity": "IGNORE"
       },
       "type": "Return"
     },
--- a/include.xml
+++ b/include.xml
@@ -4,7 +4,5 @@
   <haxelib name="thx.semver" />
 
   <haxeflag name="--macro" value="polymod.hscript._internal.HScriptRedirectDefines.run()" if="POLYMOD_REDIRECT_HSCRIPT" />
-  <!--
-    <haxeflag name="- -macro" value="polymod.hscript._internal.PolymodFinalMacro.locateAllFinals()" />
-  -->
+  <haxeflag name="--macro" value="polymod.hscript._internal.PolymodFinalMacro.locateAllFinals()" />
 </project>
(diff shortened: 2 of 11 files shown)


EliteMasterEric:
Oh yeah if you're wondering why it's a draft right now it's because it doesn't fucking work

EliteMasterEric:
image

It works now!

I mainly wrote this to help clean up FunkinCrew/Funkin#7238, so we can rewrite the variables to recycle them without the scripts being able to edit them from the outside.

Definitely should get some review from contributors but this should prevent people from accidentally messing with constants when they aren't supposed to.

An aside, this should be good to merge as-is, but in a future PR it might be nice to add functional metadata that, when applied to a class/method/block, skips this check (in a similar manner to `@:privateAccess` in Haxe source).

NotHyper-474:
On the topic of contributors reviewing, I've noticed the GCC C++ compiler would baloon in memory usage then die with this PR due to the generated meta code being too massive (and this, infact, would not be the first time this happens with this sort of macro), which prevented me from testing the code on Linux.

We likely need to rework the macro to use a Resource, akin to what https://github.com/FunkinCrew/polymod/pull/366 and some other PRs before it did.

EliteMasterEric:
> We likely need to rework the macro to use a Resource, akin to what #366 and some other PRs before it did.

I had the same issue, which is why it already uses a Resource (that probably changed between when you reviewed it and when I marked the PR as not draft).

Please review it again if you can.

NotHyper-474:
Oh, I can't believe I've missed that lol. I can confirm it compiles with no issues on my end now (besides me having to locally fix all the null-safety errors that came with the nullable `scriptInit` commit).

The only thing I found odd was that the changes from https://github.com/FunkinCrew/polymod/pull/416 were kind of reverted, was that intentional?

TechnikTil (approved):
Looks good to me!

NotHyper-474 (changes requested):
One issue I've been running with this PR are these invalid access errors that seem to specifically happen with static field access (or in this case `GameOverSubState.musicSuffix`)

Image

EliteMasterEric:
> One issue I've been running with this PR are these invalid access errors that seem to specifically happen with static field access (or in this case `GameOverSubState.musicSuffix`)
> Image

@NotHyper-474 just got this fixed let's see if we can get this merged

NotHyper-474 (approved):
Works fine now.
However, I couldn't help but notice the commit from https://github.com/FunkinCrew/polymod/pull/426, which fixed the issue, got yanked out. I suppose that was intentional considering another commit replaced it...

realvirtu:
Gotta think that this PR went from

image_2026-06-24_213744276

to

eric comment

EliteMasterEric:
> However, I couldn't help but notice the commit from #426, which fixed the issue, got yanked out. I suppose that was intentional considering another commit replaced it...

I wouldn't have made the commit to fix it if the commit from 426 had stayed in! I guess the reason the commit was gone was because of a bad rebase or force push that got rid of that change.

EliteMasterEric:
Anyways if NotHyper-474 approves it then I can merge

I should start work on the final 2.0.0 changelog soon.

TechnikTil:
> I should start work on the final 2.0.0 changelog soon.

Polymod finally gets out of `experimental`??? 👀

PR #416, opened 2026-06-11
[`final-access`] Fix unfenced import causing a compile error

NotHyper-474:
The problem was that the `polymod.util.Util` import caused a recursive access into types from OpenFL during macro time.


EliteMasterEric:
Good fix, I fixed this in OpenFL locally.

PR #420, opened 2026-06-19
chore: optimize imports, remove unused fields and correct the function return types

Starexify:
This PR does:
- optimize reports
- remove unused fields
- correct the return types from functions to use `Null<T>` instead for those that return `null`


TechnikTil:
Merge Conflicts!

PR #432, opened 2026-06-27
[BUGFIX] Make async script loading framework-agnostic

requazar:
Prior to this PR, asynchronous script loading used lime.app.Future, a class that is specific to the lime library and overall flixel backend. This PR swaps those instances out for tink.core.Future, which is framework agnostic and works with any backend.


realvirtu (approved):
I tested this before the PR was submitted, so I know this works

PR #440, opened 2026-07-02
[FIX] Only create `toString` as override if it exists in super class

NotHyper-474:
`HScriptedClassMacro` assumes the class being overriden will always have a `toString` function, which may not actually be the case.


realvirtu:
Oh my god yes!

bopel-maki-macohi:
YES

PR #441, opened 2026-07-02
Tools for fetching assets from specific mods

### LINKED ISSUE #37: [ENHANCEMENT] Add functions to retrieve files from specific mods, or to ignore mods.

EliteMasterEric:
You can use `PolymodAssets.getBytesDirectly()` to retrieve byte data directly from the `assets` folder, ignoring any mods, or use `PolymodAssets.getBytesFromMod()` to retrieve byte data directly from the mod with the given ID. The latter should work if the mod is or is not loaded.

This directly resolves #37, over 400 issues/PRs and almost 5 years later lol.


realvirtu:
image

PR #444, opened 2026-07-04
Tools for fetching assets from specific mods, additional ZIP parser fixes (for real this time)

Hundrec:
Like #441 and #443, but real


NotHyper-474:
Wowsers Hungrec bencame a Polymoder!!

That's Poly-more.

PR #448, opened 2026-07-07
Fix ambiguity with Firetongue and Polymod's `Framework` 

NotHyper-474:
Both libraries contain a `Framework` enum, which was causing a compile error due to the name ambiguity.

### DIFF
--- a/polymod/PolymodAssets.hx
+++ b/polymod/PolymodAssets.hx
@@ -1,19 +1,6 @@
 package polymod;
 
-import openfl.media.Sound;
-import haxe.io.Bytes;
 import polymod.Polymod.Framework;
-import polymod.Polymod.FrameworkParams;
-import polymod.Polymod.PolymodErrorCode;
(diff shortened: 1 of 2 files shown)


polymod/PolymodAssets.hx
```diff
@@ -1,19 +1,6 @@
...
-import polymod.Polymod.PolymodErrorCode;
-import polymod.backends.IBackend;
-import polymod.backends.PolymodAssetLibrary;
-import polymod.format.ParseRules;
-import polymod.fs.PolymodFileSystem.IFileSystem;
-import polymod.fs.StubFileSystem;
-import polymod.fs.SysFileSystem;
-#if firetongue
-import firetongue.FireTongue;
-#end
```
NotHyper-474, line 16: In this case, the ambiguity here came from the import implicitly including it. Most of the imports weren't actually being used by the class, so I went ahead and removed them as well.

PR #449, opened 2026-07-09
[ENHANCEMENT] The `BaseFileSystem` class

NotHyper-474:
This PR adds a base class for file systems to extend, providing shared implementations of methods.
It is an abstract class, meaning it cannot be instantiated directly; therefore, it can only be extended by other classes.

Since most functions have an implementation provided by this class, it makes it easier to make changes to existing FileSystems, reducing the chances of human error or the need to copy-and-paste code.

### DIFF
--- a/polymod/backends/OpenFLWithNodeBackend.hx
+++ b/polymod/backends/OpenFLWithNodeBackend.hx
@@ -319,7 +319,7 @@ class OpenFLNodeModLibrary extends LimeModLibrary
     {
       return checkIfAssetIsCached(id, type);
     }
-    else if (hasFallback)
+    else if (fallback != null)
     {
       return fallback.isLocal(id, type);
     }
@@ -335,7 +335,7 @@ class OpenFLNodeModLibrary extends LimeModLibrary
     {
       modText = cachedText.get(id);
     }
-    else if (hasFallback)
+    else if (fallback != null)
     {
       var path:String = fallback.paths.get(id);
       // check the file name for '?' and remove anything after it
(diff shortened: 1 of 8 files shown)


EliteMasterEric:
This is a great idea. This wasn't done earlier because the previous iteration of FileSystem had no shared functions, but the new one does, so this change is helpful.

NotHyper-474:
I'm keeping this a draft until I figure out why this is breaking mods with the internal code for Funkin'. Oddly, with the public one it works completely fine.

Meanwhile, I've been thinking the name `FileSystemBase` would better convey the use for this class.

realvirtu:
How is it breaking Funkin'?

NotHyper-474:
> How is it breaking Funkin'?

Mods are recognized by the game, but they aren't being able to add any songs, characters or scripts.

NotHyper-474:
Fixed it! It turned out to me a very silly oversight on my part.
I've only done some basic testing, but everything seems to work as intended now.

EliteMasterEric (changes requested):
Aside from the merge conflict present here, I'd recommend renaming the class to `BaseFileSystem` to match the other class names. Additionally, since every file system implements it anyway, we could probably remove `IFileSystem` entirely at some point, although that is a breaking change.

EliteMasterEric (approved):
Just looked over this and the code looks great, just needs a rebase to the latest `develop`.

NotHyper-474:
> Just looked over this and the code looks great, just needs a rebase to the latest `develop`.

Rebased it!

TechnikTil (approved):
`BaseFileSystem` makes me whatthehappy

PR #450, opened 2026-07-11
Improve the performance of `PolymodFinalMacro`

NotHyper-474:
Optimizes the macro to perform way less loops. While I could've also combined the loops for finals and privates, I noticed that it did not yield any significant additional improvement, and I decided to favor code readability.

Noticed an improvement from 8 to 4 seconds on my machine.
image
image


realvirtu:
jigsaw kaboom

NotHyper-474:
Just pushed a change to add caching, which improved it even more!

image

PR #455, opened 2026-07-24
Additional error reports + Add POLYMOD_STRICT_SYNTAX compiler define

nykwono:
This PR adds a couple of new error reports to scripted classes including:
- Error when trying to set a variable to be private when it's already public, vice versa.
- Error when there's an overriding function that's also static.
- Error when trying to instantiate a scripted class with a private constructor.

This also adds an additional compiler define "POLYMOD_STRICT_SYNTAX"

This compiler define is as you can guess for if the user wants scripts to have syntax handling to be more similar to Haxe, thus having it be more stricter on when errors are thrown and such. This compiler define also disallows the use of private variables completely.

Right now this define includes error reporting for:
- Having a function that overrides a super class function but doesn't use the override accessor
- Having a function with the override accessor that doesn't override anything
- Having a function with the override accessor with no parent class.
- Having a variable that's inline but also isn't static.
- Trying to retrieve a private field of a class that doesn't share inheritance. 

image

image

### DIFF
--- a/polymod/hscript/_internal/Expr.hx
+++ b/polymod/hscript/_internal/Expr.hx
@@ -160,6 +160,7 @@ enum Error
   EInvalidIterator(v:String);
   EInvalidOp(op:String);
   EInvalidAccess(f:String);
+  EPrivateField(f:String);
   EInvalidModule(m:String);
   EBlacklistedModule(m:String);
   EBlacklistedField(f:String);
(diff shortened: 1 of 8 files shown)


AbnormalPoof:
This is... so peak........

nykwono:
github doesn't want me to rebase 🥀

nykwono:
Oh yeah, will prolly bring this back to work soon and do some more with it before it's ready fro review, apologies for the hold up 😅

nykwono:
This should be good for full review now, a bit ambitious admittingly but it was my first PR lol. If this does happen to get merged I'd love to take the concept more and add support for metadata such as `@:allow` & `@:access`, `@:privateAccess`, etc.

AbnormalPoof (approved):
Wowza!

PR #459, opened 2026-08-04
[EXPERIMENTAL] Fix for ZIP files with Data Descriptors

NotHyper-474:
Should resolve https://github.com/FunkinCrew/polymod/issues/422

Makes it so the data for a LFH is retrieved from the CDH if a data descriptor is present for it, since apparently it can be used as a source of truth.


Hundrec (approved):
Absolutely works with Windows 11 zips!

PR #460, opened 2026-08-05
[BUGFIX] Fix LimeBackend's loadText not applying merges

KoloInDaCrib:
Basically just https://github.com/FunkinCrew/Funkin/pull/7842 but moved here per request.


EliteMasterEric:
Something I figured out while debugging today..,

If the game is played as a debug build with coreAssetRedirect enabled, fallback is a LimeCoreLibrary which DOES perform merging, and with coreAssetRedirect off it uses the original fallback which does NOT do merging 

Your current iteration of the PR applies append and merge TWICE in debug environments

PR #461, opened 2026-08-06
[SECURITY] Fix constructing classes ignoring blacklists.

EliteMasterEric:
This PR fixes an issue where you could construct a class by its full path regardless if it was imported, or blacklisted.

This change prevents constructing blacklisted classes but it also prevents constructing any class that isn't imported, which may negatively affect backwards compatibility.


KarimAkra (approved):
confirmed working!

PR #462, opened 2026-08-07
Use `PromiseUtil` instead of Funkin Lime's `lime.app.Promises`

nykwono:
Fixes [this issue](https://github.com/FunkinCrew/polymod/issues/457)

This PR brings Funkin Lime's `Promises.hx` into the library through `PromiseUtil.hx` and uses that instead, helping to make the library not dependent on using Funkin Lime.

### DIFF
--- a/polymod/Polymod.hx
+++ b/polymod/Polymod.hx
@@ -18,7 +18,9 @@ import thx.semver.VersionRule;
 
 using Lambda;
 using StringTools;
+#if lime
 using polymod.util.PromiseUtil;
+#end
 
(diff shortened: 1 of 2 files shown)


NotHyper-474 (commented):
Maybe those should be guarded with `#if lime` so that other backends can still work.

nykwono:
oh yeahhhh :ishowspeed:

nykwono:
kill hxFormat.json

NotHyper-474 (approved):

PR #463, opened 2026-08-08
[BUGFIX] Fix intermittent read failures on certain ZIP files

NotHyper-474:
This is a problem I've observed with QT-Rewired. Sometimes, when switching variations on Freeplay with a QT song selected there would be a random chance for the game to crash. The same could happen with stage assets, failing to load, if entering a song.

I believe the cause for this issue to be a race condition over the single FileInput handle in `ZipParser`. A handle has a pointer that dictates where it reads from, the problem is that a thread would move the pointer to a different spot while another thread was using it, causing the data to be corrupted, even though the archive itself is fine. You can imagine it as trying to mark a spot on a book then someone changing the page and causing you to mark somewhere else.

<figure>
Captura_de_tela_20260807_170007
<figcaption align="center"><sub>A local entry with a bogus file name due to the issue</sub></figcaption>

### The fix
I've decided to fix this by creating a file handle for each Local File Header instead of reusing one, which is disposed after used. I've also made sure to document it so that future code changes keep this promise. The one in `ZipParser` is still kept so that it can act as a lock. Now, you may question, why not use a mutex? A mutex would cause threads to wait for the handle to finish being used, causing parsing to be slow, since a thread would wait for the handle to be usable first. Creating a different file handle allows for concurrency without slowing things down.

### DIFF
--- a/polymod/fs/SysZipFileSystem.hx
+++ b/polymod/fs/SysZipFileSystem.hx
@@ -178,17 +178,18 @@ class SysZipFileSystem extends SysFileSystem
         return null;
       }
 
+      var fileBytes:Null<Bytes> = null;
       try
       {
-        var fileBytes = fileHeader.readData();
-        return fileBytes;
+        fileBytes = fileHeader.readData();
       }
       catch (e)
       {
         // An error occurred while reading the file from the ZIP.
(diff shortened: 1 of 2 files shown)


ahmed7p (approved):
WORKS 🧩

nykwono (approved):
Tested on QT Rewired & Sky reborn, works! (Also tested hot reloading and it works as well)

Hundrec (approved):
No more random crashes yay!

PR #473, opened 2026-08-20
Implement `interface` support

nykwono:
This PR implements full support for being able to use interfaces.

**Contents:**
- Through a macro, you're able to have any scripted classes implement **source code** interfaces (For example: `IScriptedClass` from Funkin')
- You're also able to your create own scripted interface to be used for any scripted class. 
  - With this, you're also able to extend other interfaces (including base ones!) for your interface like in regular Haxe.
- You're able to use `Std.isOfType / is` to check whether a field is a given interface (this includes fields of source code classes also, and checking for source code interfaces).
  
**Limitations:**
- Because HScript isn't typed, unfortunately interfaces aren't type checked while checking if the class satisfies it. This would be hard to do manually because of HScript's limitations of not being typed. However, most of everything else is checked for (The interface field existing, it being the right kind, etc.).

### DIFF
--- a/polymod/hscript/_internal/Expr.hx
+++ b/polymod/hscript/_internal/Expr.hx
@@ -345,9 +345,29 @@ typedef TypeDecl =
 typedef InterfaceDecl =
 {
   > ModuleType,
+
+  /**
+   * The classes imported by the scripted class
+   * This gets resolved at interpretation time to save performance and improve sandboxing
+   */
+  var imports:Map<String, ClassImport>;
+
+  /**
+   * A list of imports that have yet to be validated
+   *
+   * Scripted classes that import other scripted classes might be parsed before the class they import,
+   * so imports have to be done in two passes.
+   */
+  var importsToValidate:Map<String, ClassImport>;
+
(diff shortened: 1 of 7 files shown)


TechnikTil (approved):
whatthehappy

ahmed7p (changes requested):
When interfaces extending normal interfaces the game crashes with the lime null object window with weird bug in the terminal with out crashlog and extending ScriptedInterfaces closes the game with out any error at all like the game just close no errors no log nothing it just close + implementing does nothing it's like doesn't do anything i used it i tried to get any of the errors of not having the funtion nothing shows i tried implementing Scripted and Source ones all the same so that's what i have

i don't see any other reviews (til didn't test it) so idk if it's only me 

### some screen shots

Image

Image

Image

Image

Image

nykwono:
From the looks of it something completely broke the macro as all source code interface definitions return "null" for whatever reason now. Will look into it.

ahmed7p (approved):
Ok everything works good now

Image

AbnormalPoof (approved):
Works!

PR #480, opened 2026-08-28
Implement wildcard imports

nykwono:
This PR implements support for doing wildcard imports `(import .*)`

## How it works
Through editing the `PolymodScriptClassMacro` class, the library maps each class to its given package, with specific exclusions (Doesn't include polymod classes for obvious reasons, no generated `_Impl_` classes, etc.), this is how it works for base classes.

The import is marked to be validated later, and when the library attempts to, the wildcard import concatenates a list of classes to try importing. This importing method is the exact same as if the classes were to be regularly imported so blacklisting is still done.

These imports can also be done via an `import.hxc` file if you must.


KoloInDaCrib (approved):
Works!
Image
Image

PR #503, opened 2026-09-16
[BUGFIX] Fix default private accessor breaking scripts

nykwono:
#502 accidentally broke access modifiers and made it throw errors whenever you set a field to public, oops....... :fear:


Hundrec (approved):
lol

PR #507, opened 2026-09-18
[FEATURE] Add support for multiple compiler-metadata defines

nykwono:
This PR adds support for several compile defines including:

`@:persistent` - Stores the value of a static field even after scripts were cleared (for some reason this doesn't work sometimes which is why this PR is a draft)
`@:noUsing` - Prevents the given static function from being used when a class uses the `using` keyword on its class
`@:noPrivateAccess` - Completely prevents private access support.


polymod/hscript/_internal/Interp.hx
```diff
@@ -2315,6 +2472,16 @@ class Interp
               return obj;
             }
             return expr(e);
+          case ':noPrivateAccess':
+            // Side note: I really do not see the point of this metadata field, I guess maybe for if you're inside a private access block, but still.
```
bopel-maki-macohi, line 2476: maybe for libraries when you don't want people fucking with things they (probably) shouldn't be fucking with

nykwono:
github merge conflicts after making me have to re-code `@:using`, `@:persistent`, & `@:require` are we fucking deadass

